PR #1, opened 2022-11-23
Port extensions

jakubcolony:
This PR adds processing of extensions related events.

## Testing

This should be tested alongside [colonyCDapp#95](https://github.com/JoinColony/colonyCDapp/pull/95).

1. In the CDapp, check out the `port/42-extensions` branch. Follow the general instructions as outlined in https://github.com/JoinColony/colonyCDapp/pull/95. This will run the most up to date ingestor from this branch.
2. Play around with extensions. After every action, the UI will be updated instantly on the client-side, so you'll have to refresh the page to see if it's actually been processed by block-ingestor and stored in the DB.


.eslintrc.js
```diff
@@ -24,5 +23,6 @@ module.exports = {
     '@typescript-eslint/require-array-sort-compare': 'off',
     'no-lone-blocks': 'off',
     'no-useless-return': 'off',
+    '@typescript-eslint/member-delimiter-style': 'off',
```
jakubcolony, line 26: It was causing issues with semicolons in type/interface definitions. And its use is discouraged: https://typescript-eslint.io/rules/member-delimiter-style/

src/eventProcessor.ts
```diff
@@ -193,15 +213,103 @@ export default async (event: ContractEvent): Promise<void> => {
...
+        colony,
+      );
+
+      const receipt = await networkClient.provider.getTransactionReceipt(
+        transactionHash,
+      );
+      const installedBy = receipt.from || constants.AddressZero;
+
+      // @TODO: Get actual transaction timestamp
+      const installedAt = Math.floor(Date.now() / 1000);
```
rdig, line 244: I'm assuming this will be addressed using `getBlockTime`, right ?

src/eventProcessor.ts
```diff
@@ -193,15 +213,103 @@ export default async (event: ContractEvent): Promise<void> => {
...
+        input: {
+          item: extensionId,
+          version: BigNumber.from(version).toNumber(),
+        },
+      });
+
+      return;
+    }
+
+    case ContractEventsSignatures.ExtensionInstalled: {
```
rdig, line 229: While not really necessary, I think it would be useful to have at least some basic `verbose()` calls when handling each event.

Ie: something like `Extension <extensionAddress> has been installed in colony <colonyAddress>`

Especially since `verbose` calls will not be visible in production

rdig (approved):
Giving my approval although we already when through the changes that happened here.

Nice work!

src/eventProcessor.ts
willm30: Agree

src/eventListener.ts
```diff
@@ -1,19 +1,35 @@
...
+  await addColonyEventListener(
+    ContractEventsSignatures.ColonyFundsClaimed,
+    colonyAddress,
+  );
   await addTokenEventListener(ContractEventsSignatures.Transfer, colonyAddress);
 };
 
+/*
+ * No general, global level (basically network) events to listen to currently
+ */
```
willm30, line 24: Is this comment still relevant since we're now listening to the network events below?

willm30 (approved):
Looking good to me 👍

src/eventListener.ts
jakubcolony: Haha, you're right, I don't think I read it fully before just moving it above the handler 😅

rdig (approved):
Had a quick look over this, as I already reviewed this and it was good.

PR #3, opened 2022-12-22
Handle `ExtensionInitialised` event

jakubcolony:
This required some changes to how events are listened for in the block-ingestor and my solution feels quite hacky so @rdig feel free to suggest alternatives.

When `ExtensionInstalled` event is picked up, block-ingestor will start listening for `ExtensionInitialised` event specific to that extension instance. It will then update the extension in the database to reflect its initialised status.

For extensions that had been installed before the block-ingestor started, there's a new `trackExtensions` function for each supported extension (currently just the VotingReputation) will iterate over and sum the number of installed and uninstalled events. If the number of installations is higher than uninstallation, it will add an `ExtensionInitialised` event listener for that particular extension in a colony.

## Testing
1. Checkout CDapp to extensions branch - `git checkout port/42-extensions`.
2. Still in CDapp, apply the following patch to disable ingestor image:
```diff
diff --git a/docker/colony-cdapp-dev-env-orchestration.yaml b/docker/colony-cdapp-dev-env-orchestration.yaml
index 6e2a133..f319cc6 100644
--- a/docker/colony-cdapp-dev-env-orchestration.yaml
+++ b/docker/colony-cdapp-dev-env-orchestration.yaml
@@ -28,19 +28,19 @@ services:
     links:
     - "network-contracts:network-contracts.docker"
 
-  block-ingestor:
-    container_name: 'ingestor'
-    image: colony-cdapp-dev-env/block-ingestor:latest
-    volumes:
-      - '../amplify/mock-data:/colonyCDapp/amplify/mock-data'
-    ports:
-      - '10001:10001'
-    depends_on:
-      network-contracts:
-        condition: service_healthy
-    links:
-      - "network-contracts:network-contracts.docker"
-      - "amplify:amplify.docker"
+  # block-ingestor:
+  #   container_name: 'ingestor'
+  #   image: colony-cdapp-dev-env/block-ingestor:latest
+  #   volumes:
+  #     - '../amplify/mock-data:/colonyCDapp/amplify/mock-data'
+  #   ports:
+  #     - '10001:10001'
+  #   depends_on:
+  #     network-contracts:
+  #       condition: service_healthy
+  #   links:
+  #     - "network-contracts:network-contracts.docker"
+  #     - "amplify:amplify.docker"
 
   amplify:
     container_name: 'amplify'

```
3. Run other images with `npm run dev` in CDapp directory.
4. Start the block-ingestor with `npm run dev`.
5. Run the `temp-create-data` script. Install and enable VotingReputation extension. 
6. After refreshing the page, it should show up as Enabled:
image

7. Uninstall the extension. Restart the block-ingestor to see `trackExtensions` in action. You should see a log informing you that a listener has been added: `Added listener for Event: ExtensionInitialised()`. Enabling should still work as described above.

TODO:
 - [x] Track previously installed extensions

### DIFF
--- a/src/trackExtensions.ts
+++ b/src/trackExtensions.ts
@@ -4,6 +4,12 @@ import { extensionSpecificEventsListener } from './eventListener';
 import networkClient from './networkClient';
 import { verbose } from './utils';
 
+/**
+ * Defines extensions for which we want to track ExtensionInitialised events
+ * Note: If an extension config in CDapp doesn't specify initialization params,
+ * it is considered enabled straight after installation and we don't need
+ * to track ExtensionInitialised events for it (e.g. OneTxPayment)
+ */
 const supportedExtensionIds = [getExtensionHash(Extension.VotingReputation)];
 
 export default async (): Promise<void> => {
--- a/src/trackExtensions.ts
+++ b/src/trackExtensions.ts
@@ -1,0 +1,62 @@
+import { Extension, getExtensionHash, getLogs } from '@colony/colony-js';
+import { extensionSpecificEventsListener } from './eventListener';
+
+import networkClient from './networkClient';
+import { verbose } from './utils';
+
+const supportedExtensionIds = [getExtensionHash(Extension.VotingReputation)];
+
+export default async (): Promise<void> => {
+  verbose('Fetching already installed extensions');
+
+  supportedExtensionIds.forEach(async (extensionId) => {
+    const extensionInstalledLogs = await getLogs(
+      networkClient,
+      networkClient.filters.ExtensionInstalled(extensionId),
+    );
+    const extensionUninstalledLogs = await getLogs(
+      networkClient,
+      networkClient.filters.ExtensionUninstalled(extensionId),
+    );
+
+    /**
+     * Looping through the logs to create a mapping between colonies
+     * and the number of extension installations we encounter, minus uninstallations
+     */
+    const installedInColonyCount: { [address: string]: number } = {};
+    extensionInstalledLogs.forEach(async (log) => {
+      const parsedLog = networkClient.interface.parseLog(log);
+      const { colony } = parsedLog.args;
+
+      if (colony in installedInColonyCount) {
+        installedInColonyCount[colony]++;
+      } else {
+        installedInColonyCount[colony] = 1;
+      }
+    });
+    extensionUninstalledLogs.forEach((log) => {
+      const parsedLog = networkClient.interface.parseLog(log);
+      const { colony } = parsedLog.args;
+
+      if (colony in installedInColonyCount) {
+        installedInColonyCount[colony]--;
+      }
+    });
+
+    /**
+     * Only the colonies with installation count > 0 have this extension currently installed
+     */
+    for (const [colony, installationsCount] of Object.entries(
+      installedInColonyCount,
+    )) {
+      if (installationsCount > 0) {
+        const extensionAddress = await networkClient.getExtensionInstallation(
+          extensionId,
+          colony,
+        );
+
+        await extensionSpecificEventsListener(extensionAddress);
+      }
+    }
+  });
+};
--- a/src/trackExtensions.ts
+++ b/src/trackExtensions.ts
@@ -24,7 +24,7 @@ export default async (): Promise<void> => {
      * and the number of extension installations we encounter, minus uninstallations
      */
     const installedInColonyCount: { [address: string]: number } = {};
-    extensionInstalledLogs.forEach(async (log) => {
+    extensionInstalledLogs.forEach((log) => {
       const parsedLog = networkClient.interface.parseLog(log);
       const { colony } = parsedLog.args;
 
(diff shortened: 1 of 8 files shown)


jakubcolony:
This effectively resolves [colonyCDapp#154](https://github.com/JoinColony/colonyCDapp/issues/154) but I don't think I can link it under "Development"

src/utils.ts
```diff
@@ -206,3 +206,46 @@ export const addTokenEventListener = async (
...
+ * Extension specific event listener
+ * It creates a new interface with ABI data describing
+ * possible event signature
+ */
+export const addExtensionEventListener = async (
+  eventSignature: ContractEventsSignatures,
+  extensionAddress: string,
+): Promise<void> => {
+  const { provider } = networkClient;
+  const iface = new utils.Interface([`event ${eventSignature}`]);
```
jakubcolony, line 220: I feel this is hacky, but I didn't have a better idea. Open to suggestions on improvements

src/types.ts
```diff
@@ -36,6 +36,7 @@ export enum ContractEventsSignatures {
   ExtensionUninstalled = 'ExtensionUninstalled(bytes32,address)',
   ExtensionDeprecated = 'ExtensionDeprecated(bytes32,address,bool)',
   ExtensionUpgraded = 'ExtensionUpgraded(indexed bytes32,indexed address,uint256)',
+  ExtensionInitialised = 'ExtensionInitialised()',
```
jakubcolony, line 39: Not sure if this should belong here since it's an event specific to extension instance rather than the network (if my understanding is correct).

src/utils.ts
```diff
@@ -206,3 +206,46 @@ export const addTokenEventListener = async (
...
+    address: extensionAddress,
+  };
+
+  verbose(
+    'Added listener for Event:',
+    eventSignature,
+    extensionAddress ? `filtering Address: ${extensionAddress}` : '',
+  );
+
+  provider.on(filter, async (log: Log) => {
```
willm30, line 232: Do you know why using the networkClient interface errors when parsing the log for `ExtensionInitialised()`? I'm trying to figure it out. 

If it's because `ExtensionInitialised` is not a network-level event and therefore not in the network abi, then how come the event still gets picked up here by the network client provider?

(I'm presuming that's the reason for `addExtensionEventListener` instead of using `eventListenerGenerator`?)

src/trackExtensions.ts
```diff
@@ -0,0 +1,62 @@
+import { Extension, getExtensionHash, getLogs } from '@colony/colony-js';
+import { extensionSpecificEventsListener } from './eventListener';
+
+import networkClient from './networkClient';
+import { verbose } from './utils';
+
+const supportedExtensionIds = [getExtensionHash(Extension.VotingReputation)];
```
willm30, line 7: Should oneTxPayment be here also?

willm30 (approved):
Working for me! Left a couple of questions, really just to help me understand the block ingestor better.

src/utils.ts
jakubcolony: I have no idea. As you mentioned, `parseLog` was erroring when using the networkClient interface and I had to add the corresponding ABI to the custom interface. 
@rdig might be able to shine more light here

src/trackExtensions.ts
jakubcolony: OneTxPayment is enabled by default so we don't have to worry about its Initialised events

jakubcolony: This is determined by the lack of `initializationParams` in a particular extension config in CDapp's `constants/extensions.ts`. I believe the `OneTxPaymentInitialised` event is still possible to be emitted, but we can ignore it

rdig: Maybe leave a comment about this ?

This is kind of a _hidden_ function of the contracts, so future devs that come across this will not fall into traps.

src/trackExtensions.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export default async (): Promise<void> => {
+  verbose('Fetching already installed extensions');
+
+  supportedExtensionIds.forEach(async (extensionId) => {
```
rdig, line 12: This is going to grow super large with time, so I'm glad this will only run once at the start.

This will be a perfect candidate for when we'll start splitting up a separate "extensions"-related block ingestor

src/types.ts
rdig: Nah, don't worry about being so specific. This enum should do with all event names/sigs that we're dealing with in the block ingestor

src/utils.ts
rdig: Yeah, basically that it. The event is not present on the network contract so it doesn't know how to parse it. _(this is also because of Ethers v5 which throws instead of just returning null like the old v4)_

Jakub's solution works since he basically modified the contract _(from Ether's perspective)_

src/utils.ts
rdig: I'm ok with this, but why `Interface` and not just instantiating a new `Contract` instance ?

src/utils.ts
```diff
@@ -206,3 +206,46 @@ export const addTokenEventListener = async (
...
+    const {
+      transactionHash,
+      logIndex,
+      blockNumber,
+      address: eventContractAddress,
+    } = log;
+
+    const { hash: blockHash, timestamp } = await provider.getBlock(blockNumber);
+    addEvent({
+      ...iface.parseLog(log),
```
rdig, line 242: What happens if parsing this fails (for various other reasons) ?

I think this should be wrapped inside a try/catch or some sort of fallback

rdig (approved):
Left comments here and there, but this is very good work :100:

PR #7, opened 2023-01-12
Write extension data to the db on ingestor start

jakubcolony:
This PR adds tracking of extension events that happened before block-ingestor has started, ensuring the db stays up to date.

The tracking logic can be split in two parts: tracking current extension version and tracking extension installation events.

Latest block number refers to the latest block processed by block-ingestor that is stored in `run/stats.json`.

## Tracking current extension version
For every supported extension, we fetch the `ExtensionAddedToNetwork` events, starting from the latest block processed by the block-ingestor. It takes the most recent event (since we only care about the most recent version anyway) and if exists, writes the version to the database.

## Tracking extension installation events
For every supported extension, we fetch **all** `ExtensionInstalled` and `ExtensionUninstalled` events and then create a mapping between colony address and the count of installations minus uninstallations. If the count is 0, that means the extension has been removed at some point (we wouldn't have an entry if it was never installed), otherwise we know it's currently installed.
In case of deleted extension, we compare the most recent deletion event with the latest block number, and if it happened after that, we delete the extension from the db.
In case of installed extension, we cannot simply do that comparison. There could be many other events that happened after the installation which must be reflected in the db - deprecation, upgrading, enabling. We can get the currently installed version from the chain directly, and I added util functions for determining deprecation and initialisation status. I modified the behaviour of `updateExtensionByColonyAndHash` so that it creates an extension if one doesn't exist, and I call that mutation for each installed extension to create or update the db entry with the most recent state.

## Testing
1. (CDapp directory) Run `git checkout feat/102-track-colony-version`
2. (CDapp) Apply the following patch:
```
diff --git a/docker/colony-cdapp-dev-env-orchestration.yaml b/docker/colony-cdapp-dev-env-orchestration.yaml
index 6e2a133..f319cc6 100644
--- a/docker/colony-cdapp-dev-env-orchestration.yaml
+++ b/docker/colony-cdapp-dev-env-orchestration.yaml
@@ -28,19 +28,19 @@ services:
     links:
     - "network-contracts:network-contracts.docker"
 
-  block-ingestor:
-    container_name: 'ingestor'
-    image: colony-cdapp-dev-env/block-ingestor:latest
-    volumes:
-      - '../amplify/mock-data:/colonyCDapp/amplify/mock-data'
-    ports:
-      - '10001:10001'
-    depends_on:
-      network-contracts:
-        condition: service_healthy
-    links:
-      - "network-contracts:network-contracts.docker"
-      - "amplify:amplify.docker"
+  # block-ingestor:
+  #   container_name: 'ingestor'
+  #   image: colony-cdapp-dev-env/block-ingestor:latest
+  #   volumes:
+  #     - '../amplify/mock-data:/colonyCDapp/amplify/mock-data'
+  #   ports:
+  #     - '10001:10001'
+  #   depends_on:
+  #     network-contracts:
+  #       condition: service_healthy
+  #   links:
+  #     - "network-contracts:network-contracts.docker"
+  #     - "amplify:amplify.docker"
 
   amplify:
     container_name: 'amplify'

```
3. (CDapp) Run `npm run dev`.

With block-ingestor not running, perform different actions on extensions: install, deprecate, enable, uninstall. Run block-ingestor with `npm run dev` (in block-ingestor directory) and refresh the CDapp to verify that the results of your actions have been stored in the db successfully. Disable block-ingestor again and perform another bunch of actions, focusing on e.g. deprecating a previously installed extension. Run block-ingestor and check the results.

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -1,1 +1,9 @@
+import { Extension } from '@colony/colony-js';
+
 export const COLONY_CURRENT_VERSION_KEY = 'colony';
+
+export const SUPPORTED_EXTENSION_IDS = [
+  Extension.OneTxPayment,
+  Extension.VotingReputation,
+];
+export const INITIALISABLE_EXTENSION_IDS = [Extension.VotingReputation];
--- a/src/trackExtensions.ts
+++ b/src/trackExtensions.ts
@@ -25,10 +25,8 @@ export default async (): Promise<void> => {
     );
     await trackExtensionAddedToNetwork(extensionId, latestBlock);
 
-    verbose(
-      `Fetching events for extension: ${extensionId} starting from block: ${latestBlock}`,
-    );
-    await trackExtensionEvents(extensionId, latestBlock);
+    verbose(`Fetching events for extension: ${extensionId}`);
+    await trackExtensionEvents(extensionId);
   });
 };
 
@@ -61,10 +59,7 @@ const trackExtensionAddedToNetwork = async (
   writeExtensionVersionFromEvent(event);
 };
 
-const trackExtensionEvents = async (
-  extensionId: Extension,
-  latestBlock: number,
-): Promise<void> => {
+const trackExtensionEvents = async (extensionId: Extension): Promise<void> => {
   const extensionHash = getExtensionHash(extensionId);
   const extensionInstalledLogs = await getLogs(
     networkClient,
@@ -103,9 +98,7 @@ const trackExtensionEvents = async (
     installedInColonyCount,
   )) {
     /**
-     * If installation count is 0, that means the extension has been deleted,
-     * or never installed (but we'll try to delete it from the db anyway,
-     * to save a get query)
+     * If installation count is 0, that means the extension has been uninstalled
      */
     if (installationsCount <= 0) {
       const mostRecentUninstalledLog =
--- a/src/trackExtensions.ts
+++ b/src/trackExtensions.ts
@@ -1,36 +1,29 @@
-import { Extension, getExtensionHash, getLogs } from '@colony/colony-js';
-import { Log } from '@ethersproject/providers';
+import { getExtensionHash, getLogs } from '@colony/colony-js';
+import { BigNumber } from 'ethers';
 
 import { extensionSpecificEventsListener } from './eventListener';
 import networkClient from './networkClient';
 import {
+  isExtensionDeprecated,
   mapLogToContractEvent,
   verbose,
   writeExtensionFromEvent,
 } from './utils';
-
-/**
- * Defines extensions for which we want to track ExtensionInitialised events
- * Note: If an extension config in CDapp doesn't specify initialization params,
- * it is considered enabled straight after installation and we don't need
- * to track ExtensionInitialised events for it (e.g. OneTxPayment)
- */
-const supportedExtensionIds = [getExtensionHash(Extension.VotingReputation)];
+import { SUPPORTED_EXTENSION_IDS } from './constants';
 
 export default async (): Promise<void> => {
   verbose('Fetching already installed extensions');
 
-  supportedExtensionIds.forEach(async (extensionId) => {
+  SUPPORTED_EXTENSION_IDS.forEach(async (extensionId) => {
+    const extensionHash = getExtensionHash(extensionId);
     const extensionInstalledLogs = await getLogs(
       networkClient,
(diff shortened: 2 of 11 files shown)


src/constants.ts
```diff
@@ -0,0 +1,7 @@
+import { Extension } from '@colony/colony-js';
+
+export const SUPPORTED_EXTENSION_IDS = [
+  Extension.OneTxPayment,
+  Extension.VotingReputation,
+];
+export const INITIALISABLE_EXTENSION_IDS = [Extension.VotingReputation];
```
jakubcolony, line 7: I separated supported ids (meaning all the extensions we want to track) from initialisable ids (for which we want to add a listener for ExtensionInitialised event)

src/utils/extensions.ts
```diff
@@ -0,0 +1,166 @@
...
+import networkClient from '../networkClient';
+import { mutate } from '../amplifyClient';
+import { ContractEvent, ContractEventsSignatures } from '../types';
+import { verbose } from './logger';
+import { toNumber } from './numbers';
+
+/**
+ * Function writing the extension version to the db based on the ExtensionAddedToNetwork event payload
+ */
+export const writeExtensionVersionFromEvent = async (
```
jakubcolony, line 14: Extracted these functions into a separate file so that they can be shared by `trackExtensions` and `eventProcessor`

src/utils/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './logger';
+export * from './events';
+export * from './stats';
+export * from './extensions';
+export * from './numbers';
```
jakubcolony, line 5: `utils.ts` was growing quite large, so I split it into multiple files

src/trackExtensions.ts
```diff
@@ -1,68 +1,184 @@
...
+  writeExtensionFromEvent,
+  writeExtensionVersionFromEvent,
+} from './utils';
+import { SUPPORTED_EXTENSION_IDS } from './constants';
+import stats from '../run/stats.json';
+import { extensionSpecificEventsListener } from './eventListener';
 
 export default async (): Promise<void> => {
-  verbose('Fetching already installed extensions');
+  const latestBlock = stats.latestBlock ?? 0;
```
rdig, line 19: Theoretically you can never be at 0, as a chain starts at block 1, but this is more or less irrelevant here, as it won't really change anything

src/trackExtensions.ts
```diff
@@ -1,68 +1,184 @@
...
+   * Looping through the logs to create a mapping between colonies
+   * and the number of extension installations we encounter, minus uninstallations
+   */
+  const installedInColonyCount: { [address: string]: number } = {};
+  extensionInstalledLogs.forEach((log) => {
+    const parsedLog = networkClient.interface.parseLog(log);
+    const { colony } = parsedLog.args;
+
+    if (colony in installedInColonyCount) {
+      installedInColonyCount[colony]++;
```
rdig, line 85: I'd would have wrote this as 
```suggestion
      installedInColonyCount[colony] += 1;
```

As it makes it more readable

src/trackExtensions.ts
```diff
@@ -1,68 +1,184 @@
...
+    } else {
+      installedInColonyCount[colony] = 1;
+    }
+  });
+  extensionUninstalledLogs.forEach((log) => {
+    const parsedLog = networkClient.interface.parseLog(log);
+    const { colony } = parsedLog.args;
+
+    if (colony in installedInColonyCount) {
+      installedInColonyCount[colony]--;
```
rdig, line 95: Same here
```suggestion
      installedInColonyCount[colony] -= 1;
```

src/utils/extensions.ts
```diff
@@ -0,0 +1,166 @@
...
+    await mutate('createColonyExtension', {
+      input: {
+        id: extensionAddress,
+        ...input,
+      },
+    });
+  }
+};
+
+export const deleteExtensionFromEvent = async (
```
rdig, line 91: This name is a bit "confusing". Since while you are using a event, you're deleting the extension from the database.

But this is a nit-pick, so feel free to leave it as it is.

rdig (approved):
I really like the abstractions to write extensions from events. Very high quality work here.

I've left some comments but mostly small things, and nit-picks.

willm30 (approved):
Nice, all working for me!

PR #16, opened 2023-02-10
Add mint tokens motion support

willm30:
## Description

This pr adds block-ingestor support for the MotionCreated event, specifically when it wraps the mint tokens action.

I have coupled the MotionCreated event listeners with the Voting Reputation extension, such that as soon as it is initialised, or if `trackExtensions` determines that it is initialised on block-ingestor start up, it will also listen for MotionCreated events.

## Testing

To test, you will need to:

1. Be on this cdapp branch: https://github.com/JoinColony/colonyCDapp/pull/261. Make sure the block ingestor is commented out in in `colony-cdapp-env-orchestration.yaml`:

``` yaml

  # block-ingestor:
  #   container_name: 'ingestor'
  #   image: colony-cdapp-dev-env/block-ingestor:latest
  #   volumes:
  #     - '../amplify/mock-data:/colonyCDapp/amplify/mock-data'
  #   ports:
  #     - '10001:10001'
  #   depends_on:
  #     network-contracts:
  #       condition: service_healthy
  #     amplify:
  #       condition: service_healthy
  #   links:
  #     - "network-contracts:network-contracts.docker"
  #     - "amplify:amplify.docker"
```
2. Add reputation to the user. This can be done by:
* Copying the following into a new tab and ensuring it says the reputation miner is on: http://localhost:3001/reputation/monitor/toggle
* In a new terminal window, run `npm run truffle console`
* Paste the following:
  *  c = await IColony.at('{colonyAddress}') (where colony address is the address of the colony you're testing out)
  Then
  *  c.emitDomainReputationReward(1, accounts[0], '1000000000000000000') (make sure you're logged in with dev wallet 1).
  * Wait for the reputation mining cycle to complete and you should see the reputation update in the ui.
3. Install and initialise the Voting Reputation (Governance) extension
4. Click the "Test Mint Tokens" button. If all is set up correctly, you should be redirected to a motions screen. 

Testing should reveal that only one motion is created in the db, and it should be created in the colony you clicked "test mint tokens" in.

You should also terminate the block-ingestor and restart it, and you should see a listener attached to the MotionCreated event as a result of `trackExtensions` determining that the Voting Reputation Extn is installed.  

## Notes

In my dev env, I've been noticing that the event listener attached to the `MotionCreated` event is firing for each colony, as well as often multiple times for the same colony. 

I have guarded against the motion being associated with the wrong colony in the db, but you may see errors in the block ingestor terminal when creating a motion in the db, because it may attempt to create the same motion twice. This should not be a problem and should not actually create more than one motion per tx hash. You can verify this in the ui as well as at localhost:20002 by ensuring only one motion is created per click.
 
I have also noticed that sometimes when you click the button, it redirects but the block ingestor doesn't pick up the event and so the action page doesn't load. If this happens, go back to Colony Home and refresh & try again. We suspect this may be the result of hot reloading the block ingestor and therefore only a problem in dev. 

Finally, I've noticed that a couple of TokensMinted events get picked up straight after, with an amount of undefined and 0. I've added logic to guard against making a db mutatation when this is the case. 
  
## Troubleshooting

If you are experiencing an incorrent nonce error at colony create when running the block ingestor separately, it may be caused by the `await provider.getNetwork()` call on line 45 of index.ts. This is just a theory but commenting it out and the `getChainId()` call on the line below solved it for me the first time I tried it.


src/eventProcessor.ts
```diff
@@ -336,52 +343,94 @@ export default async (event: ContractEvent): Promise<void> => {
...
+            break;
+          }
+        }
 
-      return;
+        verbose(`${contractOperation} Motion Created`);
+      }
     }
 
+    // case ContractEventsSignatures.TokensMinted: {
```
willm30, line 398: Only commented out for testing motions. I'm noticing the TokensMinted event is firing after the CreateMotion event. A look at redux tells us the correct saga is being called (rootMotion), so not sure why this is happening.

src/eventQueue.ts
```diff
@@ -56,14 +59,21 @@ eventQueue.on(QueueEvents.QueueUpdated, function (this: EventQueue) {
...
-  verbose('Event added to the queue:', event?.signature || ContractEventsSignatures.UknownEvent);
-  this.emit(QueueEvents.QueueUpdated);
-});
+eventQueue.on(
+  QueueEvents.NewEvent,
+  function (this: EventQueue, event: ContractEvent) {
+    this.queue.push(event);
+    verbose(
+      'Event added to the queue:',
+      event?.signature || ContractEventsSignatures.UnknownEvent,
```
willm30, line 68: Changes in this file are just from prettier & correcting the "Uknown" event typo

src/types.ts
```diff
@@ -62,3 +71,33 @@ export type ChainID = number;
...
   MintTokens = 'MINT_TOKENS',
 }
+
+export enum ColonyMotionType {
+  MintTokensMotion = 'MINT_TOKENS_MOTION',
+}
+/*
+ * Contract calls
+ */
+export enum ColonyOperations {
```
willm30, line 81: A "ColonyOperation" being the underlying contract call, as distinct from a "ColonyAction" which may have multiple operations (e.g. the create colony flow)

src/utils/eventListeners.ts
```diff
@@ -0,0 +1,131 @@
...
+  eventSignature: ContractEventsSignatures,
+  contractAddress: string,
+  clientType: ClientType = ClientType.NetworkClient,
+): Promise<void> => {
+  let { client, provider } = await getClientAndProvider(
+    clientType,
+    contractAddress,
+  );
+
+  const filter = getEventFilter(eventSignature, contractAddress, clientType);
```
willm30, line 31: Have moved these to separate file to separate them from the helpers I have written for `eventListenerGenerator`

src/utils/eventListeners.ts
```diff
@@ -0,0 +1,131 @@
...
+  mapLogToContractEvent,
+} from './events';
+import { verbose } from './logger';
+/**
+ * Generator method for events listeners
+ *
+ * It basically does away with all the boilerplate of setting up topics, setting
+ * the event listener, parsing the received event
+ */
+export const eventListenerGenerator = async (
```
willm30, line 21: Have moved all the listeners to their own file to separate them from the helpers they use, especially now I've written a few for `eventListenerGenerator`

src/utils/eventListeners.ts
```diff
@@ -0,0 +1,131 @@
...
+    eventSignature,
+    contractAddress,
+    ClientType.TokenClient,
+  );
+
+/**
+ * Voting Reputation Client specific event listener,
+ * which uses `eventListenerGenerator` under the hood
+ */
+export const addMotionEventListener = async (
```
willm30, line 90: This is the only addition, aside from refactoring event listener generator

src/utils/events.ts
```diff
@@ -1,164 +1,153 @@
...
+        });
+        break;
+      }
+
+      case ClientType.VotingReputationClient: {
+        /*
+         * In the case of Voting Rep, only add the event to the queue if the colony address
+         * associated with the event is the same as the colony address the listener is associated with.
+         */
+        const { colonyId } = await query('getColonyExtension', {
```
willm30, line 128: Using this to get the colony the voting reputation extension is installed in. This is a guard against the motion being written to the db with the wrong colony (which is possible when the behaviour I'm seeing in development occurs, where the event fires on all colonies)

src/utils/motions.ts
```diff
@@ -0,0 +1,59 @@
+import { AnyColonyClient, Extension } from '@colony/colony-js';
+import { TransactionDescription } from 'ethers/lib/utils';
+
+import { mutate } from '../amplifyClient';
+import { formatDomain } from '../domains';
+import { ContractEvent, motionNameMapping } from '../types';
+import { verbose } from './logger';
+import { getColonyTokenAddress, getColonyTokenData } from './tokens';
+
+export const getParsedActionFromMotion = async (
```
willm30, line 10: This is how to get the underlying action associated with the motion, e.g mint tokens

src/utils/motions.ts
```diff
@@ -0,0 +1,59 @@
...
+    args: { creator, domainId },
+  }: ContractEvent,
+  parsedAction: TransactionDescription,
+  colonyClient: AnyColonyClient,
+) => {
+  const { name, args: actionArgs } = parsedAction;
+  const amount = actionArgs[0].toString();
+  const tokenAddress = await getColonyTokenAddress(colonyAddress);
+  const tokenData = await getColonyTokenData(tokenAddress, colonyClient);
+  await mutate('createColonyAction', {
```
willm30, line 44: This adds all the data the ui needs for the mint token motion

src/types.ts
```diff
@@ -62,3 +71,33 @@ export type ChainID = number;
...
   MintTokens = 'MINT_TOKENS',
 }
+
+export enum ColonyMotionType {
+  MintTokensMotion = 'MINT_TOKENS_MOTION',
+}
+/*
+ * Contract calls
+ */
+export enum ColonyActions {
```
jakubcolony, line 81: Could the name be more indicative that these are the contract calls?

src/types.ts
```diff
@@ -62,3 +71,33 @@ export type ChainID = number;
 export enum ColonyActionType {
   MintTokens = 'MINT_TOKENS',
 }
+
+export enum ColonyMotionType {
+  MintTokensMotion = 'MINT_TOKENS_MOTION',
+}
```
jakubcolony, line 77: I noticed in the corresponding CDapp PR you keep those merged, should we also stick to that idea here?

jakubcolony (commented):
I got it to work in conjunction with CDapp#263, but I'm only commenting for now - I had a chat with Raul today and I'll need to refactor the way we handle events that comprise an action and that will probably require some changes to this PR subsequently

src/eventProcessor.ts
```diff
@@ -360,26 +409,34 @@ export default async (event: ContractEvent): Promise<void> => {
 
       const tokenAddress = await getColonyTokenAddress(colonyAddress);
 
-      verbose(
-        amount.toString(),
-        'tokens were minted in colony:',
-        colonyAddress,
-      );
+      if (amount && amount.toString() !== '0') {
```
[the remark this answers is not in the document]
rdig: Is the motion set to very short timeout ? Ie: it finishes instantly ? Otherwise I can't really explain this.

The underlying action / event doesn't get triggered until the motion finishes (and that is done manually).

If that isn't the case I don't really understand what's going on. Maybe ask @area for his input ?

src/index.ts
```diff
@@ -61,7 +61,7 @@ const startIngestor = async (): Promise<void> => {
   /*
    * Setup all listeners we care about
    */
-  await blockListener();
```
rdig, line 64: Not saying it's something wrong with this, I'm just curious on why you removed it

src/trackExtensions.ts
```diff
@@ -138,6 +141,11 @@ const trackExtensionEvents = async (
     // Listen to extension specific events (e.g. ExtensionInitialised)
     await extensionSpecificEventsListener(extensionAddress, extensionHash);
 
+    // Listen for motions if Voting Reputation is installed.
+    if (Extension.VotingReputation === extensionId) {
+      addMotionEventListener(ContractEventsSignatures.MotionCreated, colony);
+    }
```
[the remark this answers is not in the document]
rdig: :point_up:

rdig (approved):
Left some comments, but otherwise seems good.

src/index.ts
willm30: This one's not an async function

src/utils/motions.ts
```diff
@@ -0,0 +1,58 @@
...
+) => {
+  const votingClient = await colonyClient.getExtensionClient(
+    Extension.VotingReputation,
+  );
+
+  const motion = await votingClient.getMotion(motionId);
+
+  try {
+    return colonyClient.interface.parseTransaction({
+      data: motion.action,
```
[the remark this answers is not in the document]
willm30: @rdig I don't think so, the event emits `motionId`, `creator`, and `domainId` but not the underlying action

src/types.ts
willm30: I've gone for `ColonyOperations`. Preferable?

src/trackExtensions.ts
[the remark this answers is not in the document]
willm30: Thanks, have moved down

src/utils/events.ts
```diff
@@ -1,168 +1,19 @@
...
 export const setToJS = (set: Set<string>): Array<Record<string, string>> =>
   Array.from(set).map((entry) => JSON.parse(entry));
 
-/*
- * Generator method for events listeners
- *
- * It basically does away with all the boilerplate of setting up topics, setting
- * the event listener, parsing the received event
- */
-export const eventListenerGenerator = async (
```
willm30, line 32: Moved to own file as I have added a few helpers

src/utils/events.ts
```diff
@@ -1,168 +1,19 @@
...
-      verbose('Failed to process the event: ', error);
-    }
-  });
-};
-
-/*
- * Network Client specific event listener,
- * which uses `eventListenerGenerator` under the hood
- */
-export const addNetworkEventListener = async (
```
willm30, line 97: moved to '~utils/eventListeners'

src/utils/eventListeners/eventListenerGenerator.ts
```diff
@@ -0,0 +1,177 @@
...
+  eventSignature: ContractEventsSignatures,
+  contractAddress: string,
+  clientType: ClientType = ClientType.NetworkClient,
+): Promise<void> => {
+  let { client, provider } = await getClientAndProvider(
+    clientType,
+    contractAddress,
+  );
+
+  const filter = getEventFilter(eventSignature, contractAddress, clientType);
```
willm30, line 27: Wrote some helpers for this. Should make it easier to extend.

jakubcolony:
@willm30 I pulled your latest changes and I'm seeing those errors when trying to create mint tokens motion. Looks like you're trying to write some fields (`tokenSymbol`, `fromDomain`) that have been modified since:

```
errors: [
    {
      message: 'Variable "$input" got invalid value { id: "0x97e6da42f5d3b2d5e02a7288d5b0d222322c72fe5bdfa6aba36547eeed8f868f", colonyId: "0xA4b4F5d9DBe383206Bd11C0e447F4462B9ea6cAE", type: "MINT_TOKENS_MOTION", isMotion: true, tokenSymbol: "JKB", fromDomain: "0xA4b4F5d9DBe383206Bd11C0e447F4462B9ea6cAE_1", initiatorAddress: "0xb77D57F4959eAfA0339424b83FcFaf9c15407461", amount: "1", blockNumber: 1100, tokenAddress: "0x8c60Ab84fa41a5E826BB91C30680e3b5faF8972B" }; Field "tokenSymbol" is not defined by type CreateColonyActionInput.',
      locations: [Array]
    },
    {
      message: 'Variable "$input" got invalid value { id: "0x97e6da42f5d3b2d5e02a7288d5b0d222322c72fe5bdfa6aba36547eeed8f868f", colonyId: "0xA4b4F5d9DBe383206Bd11C0e447F4462B9ea6cAE", type: "MINT_TOKENS_MOTION", isMotion: true, tokenSymbol: "JKB", fromDomain: "0xA4b4F5d9DBe383206Bd11C0e447F4462B9ea6cAE_1", initiatorAddress: "0xb77D57F4959eAfA0339424b83FcFaf9c15407461", amount: "1", blockNumber: 1100, tokenAddress: "0x8c60Ab84fa41a5E826BB91C30680e3b5faF8972B" }; Field "fromDomain" is not defined by type CreateColonyActionInput. Did you mean fromDomainId or toDomainId?',
      locations: [Array]
    }
  ]
```

willm30:
@jakubcolony Yep I will need to update this as I rebased the cdapp branch.

src/utils/tokens.ts
```diff
@@ -7,3 +9,11 @@ export const getColonyTokenAddress = async (
...
   return tokenAddress;
 };
+
+export const getColonyTokenData = async (
+  tokenAddress: string,
+  colonyClient: AnyColonyClient,
+): Promise<{ name: string; decimals: number; symbol: string }> => {
+  const tokenClient = await getTokenClient(tokenAddress, colonyClient.provider);
+  return await tokenClient.getTokenInfo();
+};
```
jakubcolony, line 19: Is this used anywhere? I can see it imported in `utils/motions` but doesn't seem to be called

jakubcolony (changes requested):
@willm30 Nice work overall, but after pulling the latest from `port/colony-motions` I can't create a colony to test this, due to the following error from `GetFullColonyByName` query:
```
"Cannot query field \"color\" on type \"Domain\". Did you mean \"colony\" or \"colonyId\"?"
```


Also, we don't have a linter yet, but do you have an eslint extension installed? I'm seeing a few eslint warnings here and there, e.g. `utils/motions.ts`:
image

willm30:
@jakubcolony I have rebased the cdapp after you rebased actions and have rebased this branch too on latest actions. I'm not experiencing problems creating a colony and all seems to have integrated correctly. But let me know if after being both branches afresh, you still have issues.

src/utils/motions.ts
```diff
@@ -0,0 +1,55 @@
...
+  {
+    transactionHash,
+    contractAddress: colonyAddress,
+    blockNumber,
+    args: { creator, domainId },
+  }: ContractEvent,
+  parsedAction: TransactionDescription,
+) => {
+  const { name, args: actionArgs } = parsedAction;
+  const amount = actionArgs[0].toString();
```
jakubcolony, line 40: Could this element of the array be null? If so, consider adding optional chaining

jakubcolony (approved):
Nice work! 💪

src/utils/motions.ts
willm30: @jakubcolony No, don't expect so. The front end should not be passing through an undefined amount of tokens to mint.

It maps to the motion params field in the current set up, for reference:

motionparams

jakubcolony: Is that a safe assumption to make, that it will never be broken by anyone?
What about calling it directly from a contract?

willm30: @jakubcolony I don't believe calling it directly from a contract will be a problem either. You need to pass an "encoded action" to the `createMotion` call, and in order for the `encode action` function to succeed, it must receive a complete set of arguments to the underlying action call (by 'complete' I mean correct length, each of which is the correct type). 

@rdig Can a malformed event find its way into the block ingestor? I.e. an event without the expected set of arguments?

jakubcolony: Good to know, thanks 👍

PR #18, opened 2023-02-16
Add motion staking support

willm30:
Adds support for motion staking.

I recommend testing with this branch in the CDapp: https://github.com/JoinColony/colonyCDapp/pull/310.

Should listen for MotionStaked events whenever it detects the VotingReputation extn is initialised and populate them to the db for use in the client.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
   "dependencies": {
     "@colony/colony-js": "^6.1.0",
     "aws-amplify": "^4.3.43",
+    "decimal.js": "^10.4.3",
     "dotenv": "^16.0.3",
     "ethers": "^5.7.2",
     "express": "^4.18.2",
--- a/src/handlers/extensions/extensionInitialised.ts
+++ b/src/handlers/extensions/extensionInitialised.ts
@@ -21,7 +21,7 @@ export default async (event: ContractEvent): Promise<void> => {
     },
   });
 
-  /* Listen for motions once Voting Reputation is enabled.*/
+  /* Listen for motions once Voting Reputation is enabled. */
   if (getExtensionHash(Extension.VotingReputation) === extensionHash) {
     await motionSpecificEventsListener(colonyAddress);
 
--- a/src/handlers/extensions/extensionInitialised.ts
+++ b/src/handlers/extensions/extensionInitialised.ts
@@ -1,12 +1,9 @@
 import { Extension, getExtensionHash } from '@colony/colony-js';
 
 import { mutate } from '~amplifyClient';
-import { ContractEvent, ContractEventsSignatures } from '~types';
-import {
-  addMotionEventListener,
-  verbose,
-  writeVotingReputationInitParamsToDB,
-} from '~utils';
+import { motionSpecificEventsListener } from '~eventListener';
+import { ContractEvent } from '~types';
+import { verbose, writeVotingReputationInitParamsToDB } from '~utils';
 
 export default async (event: ContractEvent): Promise<void> => {
   const { contractAddress: extensionAddress } = event;
@@ -26,10 +23,7 @@ export default async (event: ContractEvent): Promise<void> => {
 
   /* Listen for motions once Voting Reputation is enabled.*/
   if (getExtensionHash(Extension.VotingReputation) === extensionHash) {
-    await addMotionEventListener(
-      ContractEventsSignatures.MotionCreated,
-      colonyAddress,
-    );
+    await motionSpecificEventsListener(colonyAddress);
 
     /* Add Voting Extension Initialisation Params to DB for access by motions */
     await writeVotingReputationInitParamsToDB(id, colonyAddress);
--- a/src/handlers/extensions/extensionInitialised.ts
+++ b/src/handlers/extensions/extensionInitialised.ts
@@ -2,7 +2,11 @@ import { Extension, getExtensionHash } from '@colony/colony-js';
 
 import { mutate } from '~amplifyClient';
 import { ContractEvent, ContractEventsSignatures } from '~types';
-import { addMotionEventListener, verbose } from '~utils';
+import {
+  addMotionEventListener,
+  verbose,
+  writeVotingReputationInitParamsToDB,
+} from '~utils';
 
 export default async (event: ContractEvent): Promise<void> => {
   const { contractAddress: extensionAddress } = event;
@@ -11,7 +15,7 @@ export default async (event: ContractEvent): Promise<void> => {
 
   const {
     data: {
-      updateColonyExtension: { extensionHash, colonyAddress },
+      updateColonyExtension: { id, extensionHash, colonyAddress },
     },
   } = await mutate('updateColonyExtensionByAddress', {
     input: {
@@ -26,5 +30,8 @@ export default async (event: ContractEvent): Promise<void> => {
       ContractEventsSignatures.MotionCreated,
       colonyAddress,
     );
+
+    /* Add Voting Extension Initialisation Params to DB for access by motions */
+    await writeVotingReputationInitParamsToDB(id, colonyAddress);
   }
 };
--- a/src/handlers/extensions/extensionInitialised.ts
+++ b/src/handlers/extensions/extensionInitialised.ts
@@ -2,10 +2,20 @@ import { Extension, getExtensionHash } from '@colony/colony-js';
 
 import { mutate } from '~amplifyClient';
 import { motionSpecificEventsListener } from '~eventListener';
+import { EventProcessorContext } from '~eventProcessor';
 import { ContractEvent } from '~types';
-import { verbose, writeVotingReputationInitParamsToDB } from '~utils';
+import {
+  verbose,
+  writeVotingReputationInitParamsToDB,
+  RemoveListener,
+} from '~utils';
 
-export default async (event: ContractEvent): Promise<void> => {
+export const EXTENSION_INITIALISED_MOTION_LISTENERS = 'extensionInitialised';
+
+export default async (
+  event: ContractEvent,
+  context: EventProcessorContext,
+): Promise<void> => {
   const { contractAddress: extensionAddress } = event;
 
   verbose('Extension with address:', extensionAddress, 'was enabled');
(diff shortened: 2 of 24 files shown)


package.json
```diff
@@ -25,6 +25,7 @@
   "dependencies": {
     "@colony/colony-js": "^6.1.0",
     "aws-amplify": "^4.3.43",
+    "decimal.js": "^10.4.3",
```
willm30, line 28: This is really useful for ensuring consistency in data precision between front end and db.

src/handlers/ motions/motionCreated.ts
```diff
@@ -12,23 +13,30 @@ export default async (event: ContractEvent) => {
...
         break;
       }
       default: {
         break;
       }
     }
-
     verbose(`${contractOperation} Motion Created`);
+
+    await motionSpecificEventsListener(colonyAddress);
```
willm30, line 38: I am only listening for motion staked events once a motion has been created. Which is fine for now, so long as you don't restart the ingestor mid-motion in testing (then you'd need to create a new motion).

However, when it comes to tracking motions, this might change.

src/types/motions.ts
```diff
@@ -0,0 +1,65 @@
+import { TransactionDescription } from 'ethers/lib/utils';
```
willm30, line 1: separate file for motion types

src/utils/motions.ts
```diff
@@ -1,30 +1,88 @@
...
+  return {
+    parsedAction,
+    motionData: {
+      motionId: motionId.toString(),
+      motionStakes: {
+        raw: {
+          yay: yay.toString(),
+          nay: nay.toString(),
+        },
+        percentage: getMotionStakePercentages(yay, nay, requiredStake),
```
willm30, line 74: I'm putting the percentage in the db so the staking banner knows whether to show itself or not (since it lives outside of the StakingWidget provider). (It doesn't really cut down on calculations on the client because we still need to calculate percentages for optimistic ui purposes.)

src/index.ts
```diff
@@ -61,7 +61,7 @@ const startIngestor = async (): Promise<void> => {
   /*
    * Setup all listeners we care about
    */
-  await blockListener();
+  blockListener();
```
rdig, line 64: Why did you remove the `await` here ?

src/trackExtensions.ts
```diff
@@ -172,6 +176,11 @@ const trackExtensionEvents = async (
       mostRecentInstalledLog,
     );
 
+    // Listen for motions if Voting Reputation is initialised.
+    if (Extension.VotingReputation === extensionId && isInitialised) {
+      await motionSpecificEventsListener(colony);
+    }
```
rdig, line 182: We should also remove the listeners, if the extensions is being uninstalled

rdig (changes requested):
So the only reason I rejected this, is because of the event listeners being left hanging if the extension gets uninstalled.

On a somewhat related note, I've seen a lot of changes pertaining the extensions. Is that something you've specifically added in this PR, or it's a bad rebase cobbling some of the work that Jakub did on extensions ?

jakubcolony (commented):
Hey @willm30, running the latest `port/298-staking-sagas`, I get redirected to the Unknown transaction page after clicking on `Test Create Payment`: 
image

I think I broke it in actions, but it's been subsequently fixed so you'd need to do a rebase:
```
Could not execute mutation "createColonyAction" {
  data: null,
  errors: [
    {
      message: 'Variable "$input" got invalid value { id: "0x250a5ce06fd38265dcf597b1a6a9ca7a9133bf0128dfd606da4d95572fea3090", colonyId: "0x5083264f2AEC01bd6cC2Fc4bC3785C2B92BcBb9A", blockNumber: 1098, createdAt: "2023-04-28T13:02:52.000Z", type: "PAYMENT", fromDomain: 1, tokenAddress: "0x8c60Ab84fa41a5E826BB91C30680e3b5faF8972B", amount: "4949999999999999999", initiatorAddress: "0xb77D57F4959eAfA0339424b83FcFaf9c15407461", recipientAddress: "0xb77D57F4959eAfA0339424b83FcFaf9c15407461", fundamentalChainId: 1 }; Field "fromDomain" is not defined by type CreateColonyActionInput. Did you mean fromDomainId or toDomainId?',
      locations: [Array]
    }
  ]
}
```

src/handlers/ motions/helpers.ts
```diff
@@ -0,0 +1,135 @@
...
+  });
+};
+
+export const getMotionSide = (vote: BigNumber) => {
+  if (vote.eq(1)) {
+    return MotionSide.YAY;
+  }
+
+  return MotionSide.NAY;
+};
```
jakubcolony, line 135: This could be written as a one-liner:
```export const getMotionSide = (vote: BigNumber) => vote.eq(1) ? MotionSide.YAY : MotionSide.NAY```

File: src/handlers/ motions/helpers.ts [hunk 8]
```diff
@@ -0,0 +1,135 @@
+import Decimal from 'decimal.js';
+import { BigNumber } from 'ethers';
+import { mutate } from '~amplifyClient';
+import { MotionQuery, MotionSide, UserStake } from '~types';
+import { convertStakeToPercentage } from '~utils/motions';
+
+const getUpdatedUserStakesAmount = (
```
jakubcolony, line 7: I'm really lacking some comments here. Maybe it's because it's all new stuff to me, but also the names of the functions are very similar so some documentation to understand what's going on (high-level) would be great

src/types/extensions.ts
```diff
@@ -0,0 +1,16 @@
...
+  colonyId: string;
+  hash: string;
+  version: number;
+  installedBy: string;
+  installedAt: number;
+  isDeprecated: boolean;
+  isDeleted: boolean;
+  isInitialized: boolean;
+  extensionConfig: VotingReputationParams | null;
+}
```
jakubcolony, line 16: This looks like a potential maintenance hell. I'd rather keep it as `any` until we get codegen in place.
If we don't update those when the fields change, the mutation will fail regardless, and I feel like all we get from it is a sense of false security

jakubcolony (changes requested):
I got it to work (your branches seem cursed with the nonce issue 😆) but I'm requesting changes due to eslint issues being present in a lot of files.

I recommend you install a linting extension to your IDE for now and I have opened an issue to track setting up automatic pre-push linting as we have in the CDapp: #31

willm30:
@jakubcolony Got that linter installed and tidied up. Rebased and not seeing the payment error. After a refresh I see: 

payment-se

Which I believe is expected behaviour right now.

src/utils/events/eventListeners/eventListenerGenerator.ts
```diff
@@ -32,9 +33,22 @@ export const eventListenerGenerator = async (
...
+  const removeListener = (): void => {
+    verbose(
+      'Removed listener for Event:',
+      eventSignature,
+      contractAddress ? `filtering Address: ${contractAddress}` : '',
+    );
+    provider.off(filter, handleEvent);
+  };
+
+  return removeListener;
```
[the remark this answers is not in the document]
rdig: Sounds good. I'll try to review it a bit later today

willm30:
Temporarily putting on hold as I need to refactor the way staking widget data is handled.

src/handlers/extensions/extensionInitialised.ts
```diff
@@ -21,9 +31,15 @@ export default async (event: ContractEvent): Promise<void> => {
...
   });
 
+  let removeMotionListeners: RemoveListener[] = [];
   /* Listen for motions once Voting Reputation is enabled. */
   if (getExtensionHash(Extension.VotingReputation) === extensionHash) {
-    await motionSpecificEventsListener(colonyAddress);
+    removeMotionListeners = await motionSpecificEventsListener(colonyAddress);
+
+    /* Store remove listener functions in context, to be called when extension is uninstalled */
+    context.removeListeners[colonyAddress] = {
```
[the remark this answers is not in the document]
jakubcolony: There's a few issues I can see with this approach:
1. Introducing extra abstraction in a form of context that is being passed around as argument, which I think is unnecessary. You can simply export/import whatever variable that holds the removers (or anything other piece of data you want to share between files). You are not restricted by React state and how it impacts the rendering.
```js
  // a.ts
  export const removers = {};
  
  // b.ts 
  import { removers } from './a';

  removers.add();
  removers.remove();
```
Obviously you can add some methods wrapping the logic.

2. I think too many pieces are currently aware of the context and event removers. `eventProcessor` doesn't need to know anything about them, and they could be isolated in their own file/module
3. It requires the listener to return a specific object, which has to be then handled properly by the extension initialised handler and extension uninstalled. Again, too many things know too much about the implementation details of this solution.

It's difficult to propose something concrete without looking into it further, but the general idea I would try to follow would be:
- Figure out a way of uniquely identifying the listeners. It seems like we want to remove the listeners for all events for a particular contract address (in this case the voting reputation client address), so maybe that is enough?
- Have an object mapping the unique identifiers to the remover functions, and some methods to add and trigger those 
- Add the remover function to the object at the "source", ie. the `eventListenerGenerator` method
- Call the trigger method passing it the unique identifier whenever you need to remove a listener

The advantages of that would be that no other function (e.g. `motionSpecificEventsListener`, `addMotionEventListener`) needs to know about the concept of removing listeners and you can encapsulate all the complexity in one place

It might not work without further adjustments but I'll be happy to jump on a call (and it'd make sense to get Raul on board too) and figure something out together

src/utils/events/eventListeners/eventListenerGenerator.ts
```diff
@@ -76,7 +90,7 @@ const getEventFilter = (
   eventSignature: string,
   contractAddress: string,
   clientType: ClientType,
-) => {
+): { topics: Array<string | null>; address?: string } => {
```
jakubcolony, line 93: I added a type for it in #40. Might be worth adding a TODO comment to update it here

willm30:
@jakubcolony Thanks, I will take a look on Monday. Just noting this is not the latest iteration (this is: https://github.com/JoinColony/block-ingestor/pull/36)

rdig:
Let me know after / if this refactored and I'll take a look / review it then

willm30:
Closing, this all got handled #36 / #37

PR #22, opened 2023-02-24
Add support for MoveFunds action

jakubcolony:
This is a parallel PR to https://github.com/JoinColony/colonyCDapp/pull/291

Please refer to that PR for testing instructions

### DIFF
--- a/src/handlers/actions/moveFunds.ts
+++ b/src/handlers/actions/moveFunds.ts
@@ -1,10 +1,58 @@
+import {
+  AnyColonyClient,
+  ColonyClientV1,
+  ColonyClientV2,
+  ColonyClientV3,
+  ColonyClientV4,
+} from '@colony/colony-js';
+import { BigNumber } from 'ethers';
+import networkClient from '~/networkClient';
 import { ColonyActionType, ContractEvent } from '~/types';
-import { writeActionFromEvent } from '~/utils';
+import { toNumber, writeActionFromEvent } from '~/utils';
+import { getDatabaseDomainId } from '~/utils/domains';
+
+/**
+ * The handler makes use of colonyClient getDomainFromFundingPot method which is only
+ * available on ColonyClientV5 and above. The following type predicate allows to check
+ * we're dealing with a client that supports this method
+ */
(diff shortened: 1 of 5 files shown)


src/handlers/actions/moveFunds.ts
```diff
@@ -0,0 +1,58 @@
...
+  const { contractAddress: colonyAddress } = event;
+  const {
+    agent: initiatorAddress,
+    token: tokenAddress,
+    amount,
+    fromPot,
+    toPot,
+  } = event.args;
+
+  const colonyClient = await networkClient.getColonyClient(colonyAddress);
```
rdig, line 38: I wonder if we could optimize the ingestor somewhat if we were to store all known instances of the colony client.

I see us re-instantiating the client for a lot of tracked events.

But I actually don't know what this impact would be real-life.

rdig (approved):
All good

src/handlers/actions/moveFunds.ts
```diff
@@ -0,0 +1,58 @@
...
+ * we're dealing with a client that supports this method
+ */
+type SupportedColonyClient = Exclude<
+  AnyColonyClient,
+  ColonyClientV1 | ColonyClientV2 | ColonyClientV3 | ColonyClientV4
+>;
+const isSupportedColonyClient = (
+  colonyClient: AnyColonyClient,
+): colonyClient is SupportedColonyClient =>
+  (colonyClient as SupportedColonyClient).getDomainFromFundingPot !== undefined;
```
willm30, line 26: alternatively: 

```js
'getDomainFromFundingPot' in colonyClient
```

Then you don't need to cast `colonyClient`

willm30 (approved):
👍👍

src/handlers/actions/moveFunds.ts
jakubcolony: I think explicit casting might be better here since it will alert you if the property you're checking against doesn't exist/got removed
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

PR #23, opened 2023-02-26
Add support for EditDomain action

jakubcolony:
This PR is parallel to https://github.com/JoinColony/colonyCDapp/pull/293

PR #29, opened 2023-03-13
Fix Payment action being picked up as Move Funds 

jakubcolony:
This was due to the `ColonyFundsMovedBetweenFundingPots` event being present in both actions. I now added a check to Move Funds handler to determine if it really is that event.

## Testing
Running block-ingestor on this branch separately (you can find the patch you need to apply to the CDapp repo in https://github.com/JoinColony/block-ingestor/pull/23), create a colony using the wizard and test the payment action. It should show as Payment action instead of Move Funds.

Resolves https://github.com/JoinColony/colonyCDapp/issues/315

### DIFF
--- a/src/handlers/actions/moveFunds.ts
+++ b/src/handlers/actions/moveFunds.ts
@@ -5,10 +5,21 @@ import {
   ColonyClientV3,
   ColonyClientV4,
 } from '@colony/colony-js';
-import { BigNumber } from 'ethers';
-import networkClient from '~/networkClient';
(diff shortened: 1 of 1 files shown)


rdig (approved):
Yep, good fix. It's just something we have to deal with, as both the Payment and Move Funds actions make use of the same event _(This is because the Payment call is actually done through the extension and underneath the hood it's actually composed of multiple calls)_

Julianahlk (approved):
Nice fix! The Payment action is now being displayed as expected.

PR #30, opened 2023-03-14
Port manage reputation action

Julianahlk:
This PR adds block-ingestor support for `ArbitraryReputationUpdate` which maps to Colony Action Types `EmitDomainReputationPenalty` and `EmitDomainReputationReward`.


src/handlers/actions/EmitDomainReputation.ts
```diff
@@ -0,0 +1,43 @@
+import { BigNumber } from 'ethers';
```
jakubcolony, line 1: This file name shouldn't be capitalised

src/handlers/actions/EmitDomainReputation.ts
```diff
@@ -0,0 +1,43 @@
...
+  const colonyDomains = await Promise.all(
+    domainAddedEvents.map(async (domain) => {
+      const domainId = domain.args.domainId.toString();
+      const { skillId } = await colonyClient.getDomain(domainId);
+      return {
+        skillId,
+        domainId,
+      };
+    }),
+  );
```
jakubcolony, line 29: An alternative solution would be to query the database for the domains we store there. I don't know which one is more performant. Just something to look out in the future

src/handlers/actions/EmitDomainReputation.ts
```diff
@@ -0,0 +1,43 @@
...
+  const actionType = isReputationChangePositive ? ColonyActionType.EmitDomainReputationReward : ColonyActionType.EmitDomainReputationPenalty;
+
+  const colonyClient = await networkClient.getColonyClient(colonyAddress);
+  const domainAddedFilter = colonyClient.filters.DomainAdded(null, null);
+  const domainAddedEvents = await getEvents(colonyClient, domainAddedFilter);
+
+  const colonyDomains = await Promise.all(
+    domainAddedEvents.map(async (domain) => {
+      const domainId = domain.args.domainId.toString();
+      const { skillId } = await colonyClient.getDomain(domainId);
```
jakubcolony, line 23: Could this loop compare `skillId` as it goes and stop if it finds the right one? It doesn't seem like we need all the domains, just the one with the matching `skillId`.

src/handlers/actions/EmitDomainReputation.ts
```diff
@@ -0,0 +1,43 @@
...
+
+  const changeDomain = colonyDomains.find((domain) =>
+    domain.skillId.eq(skillId),
+  );
+
+  await writeActionFromEvent(event, colonyAddress, {
+    type: actionType,
+    initiatorAddress,
+    recipientAddress: userAddress,
+    skillId: toNumber(skillId),
```
jakubcolony, line 39: We don't need to store `skillId` at the top level of action model. If it's needed it can be accessed via `fromDomain` field

src/handlers/actions/EmitDomainReputation.ts
```diff
@@ -0,0 +1,43 @@
...
+    amount: amount.toString(),
+    fromDomainId: getDomainDatabaseId(colonyAddress, changeDomain?.domainId ?? Id.RootDomain),
```
jakubcolony, line 41: Should we not write the action at all if we can't find the domain instead of using Root as a fallback?

jakubcolony (changes requested):
Great stuff @Julianahlk 👍 I requested some small changes but it's already looking good!

src/handlers/actions/EmitDomainReputation.ts
Julianahlk: @jakubcolony  I couldn't find a simple way to do this because of the async call inside the map. So I added a new function to handle async calls inside a find.

src/handlers/actions/EmitDomainReputation.ts
Julianahlk: @jakubcolony makes sense, logging and returning if the domain cannot be found from the skillId

src/handlers/actions/EmitDomainReputation.ts
Julianahlk: @jakubcolony do you think we should create an issue for this to investigate? or should the current solution be fine for now.

rdig (approved):
Looks all good to me now, couldn't even find something to nit-pick about.

Granted, this is after all your changes per Jakub's review.

Anyway, nice work.

src/handlers/actions/emitDomainReputation.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const isReputationChangePositive = BigNumber.from(amount).gt(0);
+
+  const actionType = isReputationChangePositive ? ColonyActionType.EmitDomainReputationReward : ColonyActionType.EmitDomainReputationPenalty;
+
+  const colonyClient = await networkClient.getColonyClient(colonyAddress);
+  const domainAddedFilter = colonyClient.filters.DomainAdded(null, null);
+  const domainAddedEvents = await getEvents(colonyClient, domainAddedFilter);
+
+  const changeDomain = await findAsyncSequential(domainAddedEvents, async (domain) => {
```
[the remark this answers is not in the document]
jakubcolony: Alternatively, for improved readability, you could extract the entire logic of finding the change domain id to a separate function

src/handlers/actions/emitDomainReputation.ts
```diff
@@ -0,0 +1,40 @@
...
+    const domainId = domain.args.domainId.toString();
+    const { skillId: domainSkillId } = await colonyClient.getDomain(domainId);
+    return domainSkillId.eq(skillId);
+  });
+
+  if (!changeDomain) {
+    verbose(
+      'Not acting upon the emitDomainReputation event as a domain matching the skillId was not found',
```
jakubcolony, line 28: Nit-picking, but the event is `ArbitraryReputationUpdate`

jakubcolony (approved):
Great stuff, everything works, I'm just a bit iffy about the new util function

src/handlers/actions/emitDomainReputation.ts
Julianahlk: @jakubcolony I called the file name `emitDomainReputation` and the ColonyActionTypes are `EmitDomainReputationReward` and `EmitDomainReputationPenalty` so I thought that was the name of the Action. Should I change the file name to `arbitraryReputationUpdate` as well?

src/handlers/actions/emitDomainReputation.ts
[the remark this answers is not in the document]
Julianahlk: @jakubcolony simplified this based on your suggestions 👍

PR #32, opened 2023-03-16
Add motion finalized support

willm30:
This PR adds support for the motion finalized event in the block ingestor. Test with: https://github.com/JoinColony/colonyCDapp/pull/350


rdig (approved):
I'm mostly good with this, however I did left some comments, which I think you should consider refactoring

src/handlers/motions/helpers.ts
```diff
@@ -18,3 +19,43 @@ export const updateMotionInDB = async (
...
+ * in a previous voting reputation installation, you will get another one after uninstalling and reinstalling
+ * the voting reputation client. So, we only want the one created most recently, since that's the one created
+ * by the currently installed version of the voting reputation extension.
+ */
+
+export const getMotionFromDB = async (
+  colonyAddress: string,
+  motionId: BigNumber,
+): Promise<MotionQuery | undefined> => {
+  const { items: motions }: { items: MotionQuery[] } = await query(
```
jakubcolony, line 34: You could make `query` accept a generic that would describe the shape of the response

src/handlers/motions/index.ts
```diff
@@ -1,2 +1,3 @@
 export { default as handleMotionCreated } from './motionCreated';
 export { default as handleMotionStaked } from './motionStaked';
+export { default as handleMotionFinalized } from './motionFinalized/motionFinalized';
```
jakubcolony, line 3: Since you've got an index file in `motionFinalized` you probably don't need to double `motionFinalized` in the path?

src/handlers/motions/helpers.ts
```diff
@@ -1,6 +1,7 @@
 import { BigNumber } from 'ethers';
```
jakubcolony, line 1: Has this file got a different purpose than `utils/motions`, ie. is used by different parts of the ingestor? Otherwise I'd consider merging them or keeping them next to each other, in whatever way you think makes sense

jakubcolony (approved):
All good, left a few minor comments

PR #33, opened 2023-03-16
Add motion reward claimed support

willm30:
This PR adds block ingestor support for motion reward claimed events. Test with this branch: https://github.com/JoinColony/colonyCDapp/pull/324/ 

(note that in the cdapp branch, we actually point to https://github.com/JoinColony/block-ingestor/pull/42, which is based off this branch. This in intentional)

### DIFF
--- a/src/handlers/motions/motionRewardClaimed.ts
+++ b/src/handlers/motions/motionRewardClaimed.ts
@@ -1,0 +1,43 @@
+import { ContractEvent, MotionData } from '~types';
+import { getMotionFromDB, updateMotionInDB } from './helpers';
+
+export default async (event: ContractEvent): Promise<void> => {
+  const {
+    contractAddress: colonyAddress,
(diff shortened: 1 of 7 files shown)


src/handlers/motions/motionRewardClaimed.ts
```diff
@@ -0,0 +1,43 @@
...
+      if (address !== staker) {
+        return stakerReward;
+      }
+
+      return {
+        ...stakerReward,
+        /*
+         * This is safe because, from the front end,
+         * we claim both sides (if there were rewards on both sides)
+         * at the same time. Simply adding a flag lets us preserve the original values,
```
willm30, line 29: In other words, we don't need an isClaimed flag for each side, because we claim both at once.

rdig (approved):
Looks fine, good to go

jakubcolony (approved):
LGTM 👍

PR #35, opened 2023-03-20
Add motion data to db when motion created

willm30:
This PR adds motion data to the db at the time the `MotionCreated` event is handled.


willm30:
Have moved this commit directly to https://github.com/JoinColony/block-ingestor/pull/36, so closing.

PR #36, opened 2023-03-21
Add motion staking listeners

willm30:
This PR adds motion staking support to the block ingestor. It only handles that the event gets picked up and removed correctly. DB mutations are handled in a subsequent pr.

Test with this CDapp pr: [port/270-stake-motion-saga](https://github.com/JoinColony/colonyCDapp/pull/335) (setup instructions therein)

Also, this PR contains new functionality in the event listener generator and event processor for removing listeners. Please test this out by: 

1. installing and then immediately uninstalling the voting reputation extension
2. installing the voting rep extension; restarting the block ingestor; and then uninstalling the voting rep extn

You should see the following in the console:

rem-listene


src/eventProcessor.ts
```diff
@@ -21,10 +21,25 @@ import {
...
 } from './handlers';
+import { RemoveListener } from '~utils';
 
 dotenv.config();
 
+export enum RemoveListenerKeys {
+  EXTENSION_INITIALISED = 'extensionInitialised',
+}
+
+export interface EventProcessorContext {
```
willm30, line 34: Allows the listeners to be removed by the `extensionUninstalled` handler that were generated in `extensionInitialised`

src/utils/extensions/writeExtension.ts
```diff
@@ -163,3 +164,16 @@ export const isExtensionInitialised = async (
...
 };
+
+export const removeExtensionInitialisedListeners = (
+  colonyAddress: string,
+): void => {
+  const eventProcessContext = getEventProcessorContext();
+  const removeListeners =
+    eventProcessContext.removeListeners[colonyAddress]?.[
+      RemoveListenerKeys.EXTENSION_INITIALISED
+    ];
```
rdig, line 175: Uff. While I understand how this came about, it's really a mouth full to read :)

rdig (approved):
Seems all good.

src/utils/events/eventListeners/eventListenerGenerator.ts
```diff
@@ -32,11 +34,38 @@ export const eventListenerGenerator = async (
...
+    const listenerRemoverKey = generateListenerRemoverKey(
+      contractAddress,
+      clientType,
+      eventSignature,
+    );
+    ctx.listenerRemovers[listenerRemoverKey] = listenerRemover;
+  }
 };
 
+/* Generate a unique key for the listenerRemover. */
```
willm30, line 62: When I say "unique", I'm assuming no client type will have two different events with the exact same signature. Therefore this combination of strings will effectively prevent unintended collisions.

willm30:
@jakubcolony In this pr, I believe your concern is addressed re handling listeners. Now, the event generator listener can store the listeners in context directly, and I have a written a motion specific function that removes them and calls them (`removeMotionListeners`), which I call when the extension gets uninstalled. 

Pls review when you get a chance.

willm30:
@rdig I pushed some changes to this after you approved to take into account Jakub's suggestion of reducing the points of contact with the context object in the code base. It all stills works but if you'd like to have another glance, that'd be great

rdig (approved):
All good from my side

src/utils/motions.ts
```diff
@@ -53,3 +62,34 @@ export const writeMintTokensMotionToDB = async (
     },
   });
 };
+
+const getMotionListenerRemovers = (
```
jakubcolony, line 66: Should they be motion specific, or contract address specific? If an extension is uninstalled, a colony deleted (for example, can't happen right now), etc., we probably want to remove all the listeners for that address, not just motions?
This would be another good reason to extract the entire listener removers into their own little module

src/utils/events/eventListeners/eventListenerGenerator.ts
jakubcolony: If we kept a mapping between contract address and event signature, we could check for existing listeners before adding a new one. I experienced these collisions before in dev, because when restarted, the dev environment replays all the events that happened from block 1.

src/utils/events/eventListeners/eventListenerGenerator.ts
```diff
@@ -32,11 +34,38 @@ export const eventListenerGenerator = async (
...
+    provider.off(filter, handleEvent);
+  };
+
+  if (saveRemoverInContext) {
+    const listenerRemoverKey = generateListenerRemoverKey(
+      contractAddress,
+      clientType,
+      eventSignature,
+    );
+    ctx.listenerRemovers[listenerRemoverKey] = listenerRemover;
```
jakubcolony, line 58: I can imagine there could be a `registerListener` method that would accept some parameters, then generate the key and add it to the "cache" outside of this function

jakubcolony (changes requested):
Looks much better already but I'm asking for one more round of simplification and a few changes. 

One issue is that this is too specific for motions, and without jumping into the future, it doesn't remove the existing `ExtensionInitialised` listeners when an extension gets removed. Could some changes be made so that it removes all the listeners for a particular contract address when it's no longer relevant?

I think it's important for a new functionality like that to be as isolated as possible and only communicate with the "outside" world by some explicitly exposed methods that can hide all the complexity. Think of it as a class that could do all the work privately and only have a few public methods, such as `registerRemover`, `removeListener`, etc. Ideally, all the code related to removing listeners could form a little module within the ingestor.

src/utils/events/eventListeners/eventListenerGenerator.ts
willm30: How could this logic be moved outside of the function? We need a reference to the exact event handler we're removing. Previously I accessed this by returning it from the `eventListenerGenerator` but based on your feedback have moved it inside.

src/utils/events/eventListeners/eventListenerGenerator.ts
willm30: When restarting, memory is destroyed and recreated, no? We'd be starting with a fresh context object.

willm30:
@jakubcolony I don't in principle disagree with your feedback, but I would point out that there are many block-ingestor concepts that could have been implemented via classes, and haven't been. Given that, I feel like:

a) my proposed changes are consistent with the style of the repo, and
b) I would benefit from further reading to get myself up to speed with the pattern you're suggesting before implementing it (recommended reading welcome). 

I'd also ask you to apply the following test taken from the google engineering practices doc:

> In general, reviewers should favor approving a CL once it is in a state where it definitely improves the overall code health of the system being worked on, even if the CL isn't perfect.

I feel like your suggestions could be the subject of a refactoring issue as/when we introduce a new use case for listener removers (currently motions are the only one, right?). Else we risk running the gauntlet of YAGNI / premature abstraction.

I think the listener-remover changes made here are: 
- consistent with the repo style, and
- improve the health of the repo

They may not be perfect, but I'd prefer we handle changes in a specific refactoring issue if/when it becomes apparent they're needed.

(Btw, I do appreciate your comments! The above is more a discussion of process rather than substance.)

src/utils/events/eventListeners/eventListenerGenerator.ts
jakubcolony: What I imagined was something like this:

```js
// listenerRemover.ts 
const removers = {}; // don't need context, this can be internal to the module

export const saveRemover = (contractAddress, eventSignature, listenerRemover) => {
  const key = generateListenerRemoverKey(...)
  removers[key] = listenerRemover;
}
```

Then your code in the generator shortens to declaring the remover and calling the relevant function. There's no need for the `eventListenerGenerator` to know about how the removers are stored/accessed

jakubcolony:
@willm30 Sorry, my comment might have been written badly. What I meant is collocating the related code together (whatever style you write it; functional as you pointed out is the prevalent one) and gave an OOP class concept as one nicely depicting that - my bad, should've made it clearer. My main point was to keep it together, rather than having the logic spanning across multiple files.

> I feel like your suggestions could be the subject of a refactoring issue as/when we introduce a new use case for listener removers (currently motions are the only one, right?). Else we risk running the gauntlet of YAGNI / premature abstraction.

There's the ExtensionInitialised event listener added for each installed extension that should also be removed. There could be a separate issue for that, but I think keeping that in mind might help you refactor this easier (having a separate "module" for removing listeners that would hold all the logic). Tying it closely to motions makes it pretty difficult to refactor down the line.

willm30:
@jakubcolony Thanks for clarifying. Please review https://github.com/JoinColony/block-ingestor/pull/36/commits/13f757723c383fcb4bfbbafe1d787b87b550227f when you get a chance. 

Removing the extension initialised listener can be handled here: https://github.com/JoinColony/block-ingestor/issues/49

It should be straightforward enough and has been kept in mind when writing the functionality inside `listenerRemover.ts`.

src/utils/events/eventListeners/listenerRemover.ts
```diff
@@ -0,0 +1,68 @@
...
+const removeListeners = (listenerRemovers: ListenerRemover[]): void => {
+  listenerRemovers.forEach((listenerRemover) => {
+    listenerRemover();
+  });
+};
+
+const getRemovers = (keys: string[]): ListenerRemover[] => {
+  const listenerRemovers: ListenerRemover[] = [];
+
+  keys.forEach((key) => {
```
willm30, line 39: I could `map` here but this is slightly safer before it guards against the scenario where `removers[key]` returns undefined

willm30: Very much a just in case, not expecting it to return undefined

jakubcolony (approved):
Thanks for addressing the feedback, great work! 👌

PR #37, opened 2023-03-21
Add db mutations for motion staking

willm30:
This PR handles the database mutation for the MotionStaked event.

It introduces a number of helpers in order to calculate the latest staking data, for consumption by the staking widget in the front end, and then mutates the motionData field of the corresponding ColonyAction.

Test with [port/270-update-ui-after-staking](https://github.com/JoinColony/colonyCDapp/pull/338)

### DIFF
--- a/src/handlers/motions/motionStaked/helpers.ts
+++ b/src/handlers/motions/motionStaked/helpers.ts
@@ -1,7 +1,34 @@
 import { BigNumber } from 'ethers';
+import { MotionStakes } from '~types';
 
 export const getRequiredStake = (
   skillRep: BigNumber,
   totalStakeFraction: BigNumber,
 ): BigNumber =>
   skillRep.mul(totalStakeFraction).div(BigNumber.from(10).pow(18));
+
+export const getMotionStakes = (
+  requiredStake: BigNumber,
+  [totalNayStakesRaw, totalYayStakesRaw]: [BigNumber, BigNumber],
+): MotionStakes => {
+  const totalYayStakesPercentage = BigNumber.from(totalYayStakesRaw)
+    .div(requiredStake)
+    .mul(100);
+
+  const totalNayStakesPercentage = BigNumber.from(totalNayStakesRaw)
+    .div(requiredStake)
+    .mul(100);
+
+  const motionStakes: MotionStakes = {
+    raw: {
+      yay: totalYayStakesRaw.toString(),
+      nay: totalNayStakesRaw.toString(),
+    },
+    percentage: {
+      yay: totalYayStakesPercentage.toString(),
+      nay: totalNayStakesPercentage.toString(),
+    },
+  };
+
+  return motionStakes;
+};
--- a/src/handlers/motions/motionStaked/helpers.ts
+++ b/src/handlers/motions/motionStaked/helpers.ts
@@ -48,6 +48,37 @@ const getStakePercentage = (
   requiredStake: BigNumber,
 ): BigNumber => stake.div(requiredStake).mul(100);
 
+/**
+ * Given staking data, format and return new UserStakes object
+ */
+const getNewUserStakes = (
+  staker: string,
+  vote: BigNumber,
+  amount: BigNumber,
+  requiredStake: BigNumber,
+): UserStakes => {
+  const invertedVote = BigNumber.from(1).sub(vote);
+  const stakedSide = getMotionSide(vote);
+  const unstakedSide = getMotionSide(invertedVote);
+  return {
+    address: staker,
+    stakes: {
+      raw: {
+        [stakedSide]: amount.toString(),
+        [unstakedSide]: '0',
+      },
+
+      percentage: {
+        [stakedSide]: getStakePercentage(amount, requiredStake).toString(),
+        [unstakedSide]: '0',
+      },
+    }, // TS doesn't like the computed keys, but we know they're correct
+  } as unknown as UserStakes;
+};
+
+/**
+ * Given staking data, add stakes to existing stakes and return new, updated UserStakes object
+ */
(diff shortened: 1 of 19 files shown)


src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -138,3 +138,22 @@ export const getUpdatedUsersStakes = (
...
   return [...usersStakes, newUserStake];
 };
+
+/*
+ * Motion ids are not unique between voting reputation installations. If you had a motion with id of 1
+ * in a previous voting reputation installation, you will get another one after uninstalling and reinstalling
+ * the voting reputation client. So, we only want the one created most recently, since that's the one created
+ * by the currently installed version of the voting reputation extension.
+ */
+
```
willm30, line 148: There is at least one other way this could be handled that comes to mind. 

If we removed all motions from the db that had been initiated by a voting rep extn when it gets uninstalled, we wouldn't have this problem. 

But this means knowing the address of the voting reputation contract that initiated the motion at the time of creating the motion. To my knowledge, this data isn't present on the MotionCreated event, and therefore we'd need to store it somewhere in the block ingestor, either in memory (like in the context object I'm using for listener-removers) or in a file (like the stats file).

I will leave as is unless anyone thinks I should look into the aforementioned alternative, or something else altogether.

src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -0,0 +1,183 @@
...
+        [unstakedSide]: '0',
+      },
+    }, // TS doesn't like the computed keys, but we know they're correct
+  } as unknown as UserStakes;
+};
+
+/**
+ * Given staking data, add stakes to existing stakes and return new, updated UserStakes object
+ */
+const getUpdatedUserStakes = (
```
willm30, line 84: These two functions `getNewUserStakes` and `getUpdatedUserStakes` feel to me like really they're doing the work a good `class` should do. But we don't seem to write classes generally and I'm happy to leave as is if no objections.

src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -0,0 +1,209 @@
...
+        new Date(a.createdAt).valueOf() - new Date(b.createdAt).valueOf(),
+    );
+  }
+
+  const stakedMotion = motionsWithCorrectId.pop();
+
+  if (!stakedMotion) {
+    verbose(
+      'Could not find the staked motion in the db. This is a bug and needs investigating.',
+    );
```
willm30, line 205: We should always be able to find the motion that is being staked in the db (because it gets put there when it gets created). 

So this should never happen. Leaving a log for debugging purposes in the event it does.

src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -0,0 +1,209 @@
+import { BigNumber } from 'ethers';
+import { MotionQuery, MotionStakes, UserStakes } from '~types';
+import { verbose } from '~utils';
+import { getMotionSide } from '../helpers';
+
+export const getRequiredStake = (
+  skillRep: BigNumber,
+  totalStakeFraction: BigNumber,
+): BigNumber =>
+  skillRep.mul(totalStakeFraction).div(BigNumber.from(10).pow(18));
```
rdig, line 10: Tokens can have a different number of decimals, rather than the standard 18 as we have for most tokens

src/handlers/motions/motionStaked/helpers.ts
rdig: We do write them on occasions.

I personally don't think these two should be a class, and they can operate as-is, but if you think differently, and have a good argument for it, I'm good if you re-write them as a Class.

src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -0,0 +1,209 @@
...
+): string => {
+  /*
+   * Both totalStakeFraction and userMinStakeFraction both divide by 10 to the power of 18.
+   * Since we've multiplied by both, we need to divide by 10 to the power of 18, twice.
+   */
+
+  return skillRep
+    .mul(totalStakeFraction)
+    .mul(userMinStakeFraction)
+    .div(BigNumber.from(10).pow(36))
```
rdig, line 174: I think I even mentioned this first, when this was originally implemented in the Dapp. This can also deal with tokes that don't have 18 points of decimals, so hard-coding 36 here is not correct.

Please confirm this with Alex though, since he's the original source for this piece of logic.

src/handlers/motions/motionStaked/helpers.ts
rdig: > If we removed all motions from the db that had been initiated by a voting rep extn when it gets uninstalled, we wouldn't have this problem.

This is no longer the case using the db:
- the actual db id of the Motion is composed of the chain id, extension address and native motion id, so this will be unique across installations
- another thing I wanted to solve with a db, is to not "loose" the "old" motions once a extension instance gets uninstalled, so what you're proposing would prevent that

rdig (approved):
I left some comments, especially related to token decimals, but otherwise this looks all good.

src/handlers/motions/motionStaked/helpers.ts
willm30: Confirmed with Alex that token decimals are useful for us when displaying token data to a user but don't affect the way get required stake or min user stake are calculated

Julianahlk (approved):
Working well 👍

willm30:
Will merge once https://github.com/JoinColony/block-ingestor/pull/36 is approved

jakubcolony (approved):
LGTM, nice work! 👍

PR #38, opened 2023-03-23
Cache colony clients

### LINKED ISSUE #25: Investigate caching colony clients optimisation opportunity
>I wonder if we could optimize the ingestor somewhat if we were to store all known instances of the colony client.

>I see us re-instantiating the client for a lot of tracked events.

>But I actually don't know what this impact would be real-life.

_Originally posted by @rdig in https://github.com/JoinColony/block-ingestor/pull/22#discussion_r1118176579_
            
This issue is to investigate and work on or close the ticket.

jakubcolony:
This PR adds caching of colony clients. It was originally suggested by Raul in one of my PRs and it made sense to do it now as I found myself calling `getColonyClient` in different methods for the same colony

Resolves #25

### DIFF
--- a/src/utils/colonyClient.ts
+++ b/src/utils/colonyClient.ts
@@ -1,0 +1,21 @@
+import { AnyColonyClient } from '@colony/colony-js';
+
+import networkClient from '~networkClient';
+
+const colonyClientsCache: Record<string, AnyColonyClient> = {};
+
+/**
+ * Method attempting to return a cached colony client,
+ * or get one from the networkClient if there's no cached entry
+ */
+export const getCachedColonyClient = async (
+  colonyAddress: string,
+): Promise<AnyColonyClient> => {
+  if (colonyAddress in colonyClientsCache) {
+    return colonyClientsCache[colonyAddress];
+  }
+
+  const colonyClient = await networkClient.getColonyClient(colonyAddress);
+  colonyClientsCache[colonyAddress] = colonyClient;
+  return colonyClient;
+};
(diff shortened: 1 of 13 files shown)


Julianahlk (approved):
Looks good. Nice optimization 👍

src/utils/colonyClient.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import networkClient from '~networkClient';
+
+const colonyClientsCache: Record<string, AnyColonyClient> = {};
+
+/**
+ * Method attempting to return a cached colony client,
+ * or get one from the networkClient if there's no cached entry
+ */
+export const getCachedColonyClient = async (
```
rdig, line 11: This is a very good start and I like your approach.

But for future improvements I would extend this to also be able to fetch/cache all types of contract clients: network, colony, tokens, but more important extensions.

rdig (approved):
Yup. This is a very nice improvement.

If this was already in production, I suspect this change would have been felt in response times.

src/utils/colonyClient.ts
jakubcolony: Thanks for the suggestions. I thought about doing it in this PR, but here's what I'm proposing we do instead:
- Network client - for now, it looks like `networkClient.ts` instantiates the client once and exports the same instance everywhere; This will only need to be addressed when we support multi-chain
- Token client - opened #43, I know Will is refactoring how we get different types of client and I don't want to make his life harder by introducing more changes
- Extensions client - opened #44, this is "blocked" by Will's motion work (nothing else uses extension clients)

Issue #39, opened 2023-03-23
Optimize extensions/actions tracking by starting at the last processed block number


jakubcolony:
Currently, block-ingestor goes through all events starting from block number 1 to fetch and write past extensions/actions related data. This can (and should) be improved by starting the scanning at the last processed block number before the block-ingestor went down.

PR #40, opened 2023-03-23
Track past actions 

jakubcolony:
This PR adds tracking of past actions that might have happened when block-ingestor was down.

Similarly to extensions tracking, it currently goes through all events since block number 1 and I opened an issue to change it in the future: #39 

## Testing
Testing involves triggering all the supported actions while block-ingestor is disabled (so you need to start the CDapp on `master` without it running), then running the ingestor separately and verifying that the actions have been picked up and written to the database.

### DIFF
--- a/src/trackColonyActions.ts
+++ b/src/trackColonyActions.ts
@@ -3,6 +3,7 @@ import { utils } from 'ethers';
 
 import {
   handleMintTokensAction,
+  handleMoveFundsAction,
   handlePaymentAction,
   handleTokenUnlockedAction,
 } from '~handlers';
@@ -56,5 +57,10 @@ export default async (colonyAddress: string): Promise<void> => {
     ContractEventsSignatures.TokenUnlocked,
     colonyAddress,
     handleTokenUnlockedAction,
+  );
+  await trackActionsByEvent(
+    ContractEventsSignatures.ColonyFundsMovedBetweenFundingPots,
+    colonyAddress,
+    handleMoveFundsAction,
   );
 };
--- a/src/trackColonyActions.ts
+++ b/src/trackColonyActions.ts
@@ -1,0 +1,25 @@
+import { getLogs } from '@colony/colony-js';
+import { utils } from 'ethers';
+
+import { handleMintTokensAction } from '~handlers';
+import networkClient from '~networkClient';
+import { ContractEventsSignatures } from '~types';
+import { mapLogToContractEvent, verbose } from '~utils';
+
+export default async (colonyAddress: string): Promise<void> => {
+  verbose('Fetching past actions for colony:', colonyAddress);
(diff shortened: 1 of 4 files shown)


src/trackColonyActions.ts
```diff
@@ -0,0 +1,92 @@
...
+});
+
+const trackActionsByEvent = async (
+  eventSignature: ContractEventsSignatures,
+  colonyAddress: string,
+  handler: ColonyActionHandler,
+): Promise<void> => {
+  const colonyClient = await getCachedColonyClient(colonyAddress);
+  const filter = getFilter(eventSignature, colonyAddress);
+  const logs = await getLogs(networkClient, filter);
```
rdig, line 34: While you can't really do anything about it in here to fix it, I think a comment needs to be left here for others who stumble on this to be aware:

This can only work with an archive node, as well as one that can display more than 10000 events. Most commercial solutions out there limit to around 10k events in the past, plus that not everyone will give up an archive node.

So this can only work with our custom, nethermind based node.

rdig (approved):
I think a comment needs to be added to the main method about the supported node types, but otherwise, this is all good.

src/utils/events.ts
```diff
@@ -42,13 +42,13 @@ export const eventListenerGenerator = async (
...
 
-  const filter: { topics: Array<string | null>; address?: string } = {
+  const filter: Filter = {
     topics: [utils.id(eventSignature)],
   };
 
   if (clientType === ClientType.TokenClient) {
     filter.topics = [
-      ...filter.topics,
+      ...(filter.topics ?? []),
```
willm30, line 51: nitpick: this looks a little odd given we know topics is an array here (line 46). But I suppose Filter is telling us that topics might be undefined. 

Maybe this is a good use case for the non-null assertion `!`?

willm30 (approved):
LGTM and I will be using this as the basis for the same motions functionality!

src/utils/events.ts
jakubcolony: So the reason for that is because `filter` in line 45 is defined as of type `Filter`, which says the `topics` field can be undefined/null. At the same time, as you pointed out we know it's definitely not undefined.

The best solution would be to use the `satisfies` operator (https://devblogs.microsoft.com/typescript/announcing-typescript-4-9-beta/#the-satisfies-operator) introduced in TS 4.9 (we're on 4.8.x). It allows you to define the constraint for a value (it has to match the `Filter` type), but at the same time allows TS to infer the more precise type (since we're defining the `topics`, they can't be undefined/null):

```js
const filter = {
  topics: [utils.id(eventSignature)],
} satisfies Filter;
```

Until we upgrade, I think redundancy is safer than assertion (if I used `!` operator, changing topics to null in line 46 would not trigger a compiler error but it would lead to a runtime TypeError).

PR #42, opened 2023-03-31
Make unique database motion id

willm30:
This pr updates the relevant motion handlers to fetch the motion from the db using the unique database id, as opposed to using the native id and sorting by created date.

### DIFF
--- a/src/handlers/motions/helpers.ts
+++ b/src/handlers/motions/helpers.ts
@@ -20,12 +20,11 @@ export const updateMotionInDB = async (
   });
 };
 
-/*
- * Motion ids are not unique between voting reputation installations. If you had a motion with id of 1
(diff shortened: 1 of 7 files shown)


src/handlers/motions/helpers.ts
```diff
@@ -38,19 +38,10 @@ export const getMotionFromDB = async (
     },
   );
 
-  const motionsWithCorrectId = motions.filter(
-    ({ motionData: { motionId: id } }) => id === motionId.toString(),
+  const motion = motions.find(
```
rdig, line 41: So I think I just left a comment about refactoring this in #32 

Feel free to ignore that then :)

rdig (approved):
Very nice. It's exactly what I imagined when writing that comment explaining this in #32

Julianahlk (approved):
LGTM

PR #45, opened 2023-04-05
Define events for motion data

Julianahlk:
This PR:
- adds events to motion data
- creates an event when a motion is created
- creates an event when a motion is staked

### DIFF
--- a/src/handlers/motions/motionCreated/helpers.ts
+++ b/src/handlers/motions/motionCreated/helpers.ts
@@ -1,12 +1,10 @@
 import { BigNumber } from 'ethers';
 import { TransactionDescription } from 'ethers/lib/utils';
 
-import { mutate } from '~amplifyClient';
 import { MotionData, ContractEvent, MotionEvents } from '~types';
-import {
-  getVotingClient,
-  verbose,
-} from '~utils';
+import { getVotingClient, verbose } from '~utils';
+
+import { mutate } from '~amplifyClient';
 
 import {
   getMotionDatabaseId,
@@ -123,9 +121,9 @@ const getMotionData = async ({
 export const createMotionInDB = async (
   {
     transactionHash,
+    blockNumber,
     logIndex,
     contractAddress: colonyAddress,
-    blockNumber,
     args: { motionId, creator: creatorAddress, domainId },
   }: ContractEvent,
   input: Record<string, any>,
--- a/src/handlers/motions/motionCreated/helpers.ts
+++ b/src/handlers/motions/motionCreated/helpers.ts
@@ -42,7 +42,7 @@ export const getMotionData = async (
   logIndex: number,
   motionId: BigNumber,
   domainId: BigNumber,
-  creator: string,
+  creatorAddress: string,
 ): Promise<MotionData> => {
   const votingClient = await getVotingClient(colonyAddress);
   const { skillRep, rootHash, repSubmitted } = await votingClient.getMotion(
@@ -98,7 +98,7 @@ export const getMotionData = async (
     },
     repSubmitted: repSubmitted.toString(),
     skillRep: skillRep.toString(),
-    messages: [{ name: 'MotionCreated', messageKey: `${transactionHash}${logIndex}`, initiatorAddress: creator }],
+    messages: [{ name: 'MotionCreated', messageKey: `${transactionHash}${logIndex}`, initiatorAddress: creatorAddress }],
   };
 };
 
--- a/src/handlers/motions/motionCreated/helpers.ts
+++ b/src/handlers/motions/motionCreated/helpers.ts
@@ -98,7 +98,7 @@ export const getMotionData = async (
     },
     repSubmitted: repSubmitted.toString(),
     skillRep: skillRep.toString(),
-    messages: [{ name: 'MotionCreated', transactionHash, logIndex, initiatorAddress: creator }],
+    messages: [{ name: 'MotionCreated', messageKey: `${transactionHash}${logIndex}`, initiatorAddress: creator }],
   };
 };
 
--- a/src/handlers/motions/motionCreated/helpers.ts
+++ b/src/handlers/motions/motionCreated/helpers.ts
@@ -1,10 +1,12 @@
 import { BigNumber } from 'ethers';
 import { TransactionDescription } from 'ethers/lib/utils';
 
-import { ContractEvent, MotionData } from '~types';
-import { getVotingClient, verbose } from '~utils';
-
 import { mutate } from '~amplifyClient';
+import { MotionData, ContractEvent, MotionEvents } from '~types';
+import {
+  getVotingClient,
+  verbose,
+} from '~utils';
 
 import {
   getMotionDatabaseId,
@@ -110,7 +112,7 @@ const getMotionData = async ({
     hasObjection: false,
     messages: [
       {
-        name: 'MotionCreated',
+        name: MotionEvents.MotionCreated,
         messageKey: `${transactionHash}${logIndex}`,
         initiatorAddress: creatorAddress,
       },
--- a/src/handlers/motions/motionCreated/helpers.ts
+++ b/src/handlers/motions/motionCreated/helpers.ts
@@ -37,12 +37,12 @@ export const getParsedActionFromMotion = async (
 };
 
 interface Props {
-  colonyAddress: string,
-  transactionHash: string,
-  logIndex: number,
-  motionId: BigNumber,
-  domainId: BigNumber,
-  creatorAddress: string,
+  colonyAddress: string;
+  transactionHash: string;
+  logIndex: number;
+  motionId: BigNumber;
+  domainId: BigNumber;
+  creatorAddress: string;
 }
 
 const getMotionData = async ({
@@ -108,7 +108,13 @@ const getMotionData = async ({
     repSubmitted: repSubmitted.toString(),
     skillRep: skillRep.toString(),
     hasObjection: false,
-    messages: [{ name: 'MotionCreated', messageKey: `${transactionHash}${logIndex}`, initiatorAddress: creatorAddress }],
+    messages: [
+      {
+        name: 'MotionCreated',
+        messageKey: `${transactionHash}${logIndex}`,
+        initiatorAddress: creatorAddress,
+      },
+    ],
   };
 };
 
@@ -122,7 +128,14 @@ export const createMotionInDB = async (
   }: ContractEvent,
   input: Record<string, any>,
 ): Promise<void> => {
-  const motionData = await getMotionData({ colonyAddress, transactionHash, logIndex, motionId, domainId, creatorAddress });
+  const motionData = await getMotionData({
+    colonyAddress,
+    transactionHash,
+    logIndex,
+    motionId,
+    domainId,
+    creatorAddress,
+  });
 
   await mutate('createColonyAction', {
     input: {
--- a/src/handlers/motions/motionCreated/helpers.ts
+++ b/src/handlers/motions/motionCreated/helpers.ts
@@ -36,14 +36,23 @@ export const getParsedActionFromMotion = async (
   return undefined;
 };
 
-export const getMotionData = async (
+interface Props {
   colonyAddress: string,
   transactionHash: string,
   logIndex: number,
   motionId: BigNumber,
   domainId: BigNumber,
   creatorAddress: string,
-): Promise<MotionData> => {
+}
+
+const getMotionData = async ({
+  colonyAddress,
+  transactionHash,
+  logIndex,
+  motionId,
+  domainId,
+  creatorAddress,
+}: Props): Promise<MotionData> => {
   const votingClient = await getVotingClient(colonyAddress);
   const { skillRep, rootHash, repSubmitted } = await votingClient.getMotion(
     motionId,
@@ -98,6 +107,7 @@ export const getMotionData = async (
     },
     repSubmitted: repSubmitted.toString(),
     skillRep: skillRep.toString(),
+    hasObjection: false,
     messages: [{ name: 'MotionCreated', messageKey: `${transactionHash}${logIndex}`, initiatorAddress: creatorAddress }],
   };
 };
@@ -112,7 +122,7 @@ export const createMotionInDB = async (
   }: ContractEvent,
   input: Record<string, any>,
 ): Promise<void> => {
-  const motionData = await getMotionData(colonyAddress, transactionHash, logIndex, motionId, domainId, creatorAddress);
+  const motionData = await getMotionData({ colonyAddress, transactionHash, logIndex, motionId, domainId, creatorAddress });
 
   await mutate('createColonyAction', {
     input: {
--- a/src/handlers/motions/motionCreated/helpers.ts
+++ b/src/handlers/motions/motionCreated/helpers.ts
@@ -38,8 +38,11 @@ export const getParsedActionFromMotion = async (
 
 export const getMotionData = async (
   colonyAddress: string,
+  transactionHash: string,
+  logIndex: number,
   motionId: BigNumber,
   domainId: BigNumber,
+  creator: string,
 ): Promise<MotionData> => {
   const votingClient = await getVotingClient(colonyAddress);
   const { skillRep, rootHash, repSubmitted } = await votingClient.getMotion(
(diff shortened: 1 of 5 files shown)


src/handlers/motions/motionCreated/helpers.ts
```diff
@@ -80,6 +80,7 @@ const getMotionData = async (
     motionDomainId: domainId.toString(),
     stakerRewards: [],
     isFinalized: false,
+    events: [{ name: 'MotionCreated', initiatorAddress: '', vote: '', amount: '' }],
```
rdig, line 83: `initiatorAddress` should have a default value of the null address

rdig (approved):
All good

src/handlers/motions/motionCreated/helpers.ts
willm30: Also, why are vote and amount present here? Neither are relevant to the motion being created.

src/handlers/motions/motionStaked/motionStaked.ts
```diff
@@ -39,6 +40,7 @@ export default async (event: ContractEvent): Promise<void> => {
       motionData: { usersStakes },
       motionData,
     } = stakedMotion;
+    motionData.events.push({ name: 'MotionStaked', initiatorAddress: staker, vote: BigNumber.from(vote).toString(), amount: BigNumber.from(amount).toString() });
```
willm30, line 43: `events` can be destructured like usersStakes

willm30: vote and amount are already BigNumbers because they're typed as `uint256`, so you don't need to create them using `BigNumber.from`

willm30: And for consistency, can we write this as:

``` js

const updatedEvents = [
 ...events,
{     
      name: 'MotionStaked',
      initiatorAddress: staker,
      vote:  vote.toString(),
      amount: amount.toString()
}
]

...

await updateMotionInDB(id, {
      ...motionData,
      usersStakes: updatedUserStakes,
      motionStakes,
      remainingStakes,
      events: updatedEvents,
    });

```

src/types/motions.ts
```diff
@@ -73,3 +74,10 @@ export interface MotionQuery {
   motionData: MotionData;
   createdAt: string;
 }
+
+export interface Event {
+  name: string;
+  initiatorAddress: string;
+  vote: string;
```
willm30, line 81: Fields that might not be on a given event should be optional

willm30: I.e vote, amount

willm30 (commented):
Nice one, left a few comments but looking good. Will review again once you've taken a look

src/handlers/motions/motionCreated/helpers.ts
Julianahlk: I am using the creator now which I believe is the same as msgSender.

Julianahlk:
@rdig @willm30  I added new events for the missing messages, ObjectionRaised, MotionFullyStaked, MotionFullyStakedAfterObjection, ObjectionFullyStaked
I noticed in the cdapp some of these are SystemMessages instead of ColonyAndExtensionsEvents. The question is if its important to keep the SystemMessages separate in which case I would add systemMessages in MotionData or if we can combine them, in which case I would combine the SystemMessages into ColonyAndExtensionsEvents in the cdapp.

src/types/motions.ts
```diff
@@ -84,3 +85,12 @@ export interface MotionQuery {
   motionData: MotionData;
   createdAt: string;
 }
+
+export interface Event {
+  name: string;
+  transactionHash: string;
+  logIndex: number;
```
Julianahlk, line 92: @rdig logIndex is always zero. Is this expected as I am just adding this to create the key in the ActionDetailsPageFeed?

src/handlers/motions/motionStaked/motionStaked.ts
```diff
@@ -36,10 +39,54 @@ export default async (event: ContractEvent): Promise<void> => {
   if (stakedMotion) {
     const {
       id,
-      motionData: { usersStakes },
+      motionData: { usersStakes, events: updatedEvents },
       motionData,
     } = stakedMotion;
 
+    if (vote.eq(0) && !updatedEvents.find(({ name }) => name === 'ObjectionRaised')) {
```
willm30, line 46: This raw 0 should use the enum, as in: `MotionVote.Nay`

src/types/motions.ts
```diff
@@ -84,3 +85,12 @@ export interface MotionQuery {
   motionData: MotionData;
   createdAt: string;
 }
+
+export interface Event {
```
willm30, line 89: @rdig Related to Juliana's question about combining events and system messages, if it's fine to have one model to capture both, I propose we rename it to `Messages`.

"Event" is ambiguous since we'd not be tracking network events exclusively, and even in that event, since we're only recording a small-subset of the actual event's data, calling it an event feels misleading. 

Really what we're capturing is the data needed to display a message to the user in the motion's page, so I feel this should drive the naming of the model.

src/handlers/motions/motionStaked/motionStaked.ts
willm30, line 46: I think some of this logic could tidied up a little bit, e.g. 

``` js
const isFirstObjection = ...

if(isFirstObjection) {

}
```

instead of 

``` js
if (vote.eq(0) && !updatedEvents.find(({ name }) => name === 'ObjectionRaised')
```

But the main thing is please move all of this to a helper and follow the pattern in this file (e.g. as with user stakes)

``` js

 if (stakedMotion) {
    const {
      id,
      motionData: { usersStakes, events },
      motionData,
    } = stakedMotion;

const updatedEvents = getUpdatedEvents(events, vote, amount)

...

 await updateMotionInDB(id, {
      ...motionData,
      usersStakes: updatedUserStakes,
      motionStakes,
      remainingStakes,
      events: updatedEvents,
    });
```

src/handlers/motions/motionStaked/motionStaked.ts
```diff
@@ -36,10 +39,54 @@ export default async (event: ContractEvent): Promise<void> => {
...
+        transactionHash,
+        logIndex,
+        initiatorAddress: staker,
+        vote: vote.toString(),
+        amount: amount.toString(),
+      },
+    );
+
+    if (requiredStake.eq(BigNumber.from(motionStakes.raw.yay))) {
+      const eventName = !updatedEvents.find(({ name }) => name === 'ObjectionRaised') ? 'MotionFullyStaked' : 'MotionFullyStakedAfterObjection';
```
willm30, line 69: A faster way to do this than to iterate through all previous messages would be to store a `hasObjection` flag on the MotionData model that you set to true in the event the `vote === MotionVote.Nay`. 

Then:  

``` js

const eventName = hasObjection  ? 'MotionFullyStaked' : 'MotionFullyStakedAfterObjection';

```

src/handlers/motions/motionStaked/motionStaked.ts
```diff
@@ -36,10 +39,54 @@ export default async (event: ContractEvent): Promise<void> => {
...
+      updatedEvents.push(
+        {
+          name: eventName,
+          transactionHash,
+          logIndex,
+          initiatorAddress: staker,
+        },
+      );
+    }
+    if (requiredStake.eq(BigNumber.from(motionStakes.raw.nay))) {
```
willm30, line 79: You don't need to create a BigNumber for the motion stake since `requiredStake` is already a BigNumber and BigNumber methods can handle strings

src/handlers/motions/motionStaked/motionStaked.ts
```diff
@@ -14,6 +15,8 @@ import {
 export default async (event: ContractEvent): Promise<void> => {
   const {
     contractAddress: colonyAddress,
+    logIndex,
```
willm30, line 18: If these are the only two components of the key, we could just declare const at the top here: 
``` js
const messageKey = `${transactionHash} ${logIndex}`
```js
And then include that on every message instead of both the tx hash and log index

Then the front end doesn't need to know anything about them, it just takes a key and use it as is.

Comment (willm30) [reply src/handlers/motions/motionStaked/motionStaked.ts:18]: And while I'm thinking about it, the order of the messages isn't going to change, right?

Therefore we don't really need the logIndex either. You could just use the item's array index when mapping through in the front end. This is [not recommended](https://react.dev/learn/rendering-lists) if the ordering might change, but since ours won't, would be one less thing to think about in the block-ingestor.

File: src/types/motions.ts (hunk 4 above)
Comment (rdig) [reply src/types/motions.ts:92]: @Julianahlk 
LogIndex is always 0 for you due to the nature of the local chain, which just sends one transaction at a time, meaning your events will always have index 0 locally, however in production this number will almost always be non-zero due to multiple transactions and events being handled inside a single block. So your combination of `<transactionHash>_<logIndex>` will always be unique, both locally and in production

@willm30 

No, since multiple events can be emmited inside a single block, meaning the IDs will be duplicated

File: src/types/motions.ts (hunk 6 above)
Comment (rdig) [reply src/types/motions.ts:89]: System Messages are generated locally based on various states of events. I don't have clear idea how to handle them now. 

From a separation concern, I'd want a separate model, from a logic reuse concern I'd want to store them in the same model. Maybe have a call about this?

At any rate, do note that System Message won't have the same values as a normal event does.

Comment (willm30) [reply src/types/motions.ts:89]: Is there another reason, aside from to display messages, to track the "events" on the motion data model? If not, I think we just call it `messages` or `systemMessages`. I.e. what would be the reason to keep them separate? 

I say "events" because we're not actually storing the full event object, we're storing a subset of its fields that are useful for displaying messages. So really it's the same concern, as far as I can tell.

File: src/handlers/motions/motionStaked/helpers.ts [hunk 10]
```diff
@@ -174,3 +174,59 @@ export const getUserMinStake = (
...
+  updatedMessages: Message[],
+  requiredStake: BigNumber,
+  motionStakes: MotionStakes,
+  transactionHash: string,
+  logIndex: number,
+  vote: BigNumber,
+  staker: string,
+  amount: BigNumber,
+): Message[] => {
+    const messageKey = `${transactionHash}${logIndex}`;
```
willm30, line 188: This can take `messageKey` as a prop instead of composing it here

src/handlers/motions/motionStaked/motionStaked.ts
```diff
@@ -48,11 +51,14 @@ export default async (event: ContractEvent): Promise<void> => {
       requiredStake,
     );
 
+    const updatedMessages = getUpdatedMessages(messages, requiredStake, motionStakes, transactionHash, logIndex, vote, staker, amount);
```
willm30: There's quite a few props here. Maybe you'd be better off having a single argument that's an object, with the current arguments as object properties. Then you could do something like:

```js
getUpdatedMessages({
  messages, 
  requiredStake, 
  motionStakes, 
  messageKey: `${transactionHash}${logIndex}`
  ...args
})
```

It's a little better and means you don't have to worry about ordering.

src/handlers/motions/motionCreated/helpers.ts
```diff
@@ -111,7 +116,7 @@ export const writeMintTokensMotionToDB = async (
   const { name, args: actionArgs } = parsedAction;
   const amount = actionArgs[0].toString();
   const tokenAddress = await getColonyTokenAddress(colonyAddress);
-  const motionData = await getMotionData(colonyAddress, motionId, domainId);
+  const motionData = await getMotionData(colonyAddress, transactionHash, logIndex, motionId, domainId, creator);
```
willm30, line 119: Consider refactoring this as well to take an object as an argument. It would make extending it easier since the argument ordering would be irrelevant.

willm30 (commented):
I think this is more or less good to go. One thing I would like to see is the `hasObjection` property added to the motionData model, since why iterate through them all just to find this out if we don't have to?

src/types/motions.ts
```diff
@@ -84,3 +85,11 @@ export interface MotionQuery {
   motionData: MotionData;
   createdAt: string;
 }
+
+export interface Message {
```
jakubcolony, line 89: Should this be called `MotionMessage` for full clarity?

jakubcolony (approved):
The ingestor part looks good 👍 I left a few comments in the corresponding CDapp PR
Nice work!

src/handlers/motions/motionCreated/helpers.ts
```diff
@@ -38,8 +38,11 @@ export const getParsedActionFromMotion = async (
 
 const getMotionData = async (
   colonyAddress: string,
+  transactionHash: string,
+  logIndex: number,
   motionId: BigNumber,
   domainId: BigNumber,
+  creator: string,
```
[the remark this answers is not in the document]
Julianahlk: @jakubcolony  How about creatorAddress?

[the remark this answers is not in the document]
jakubcolony: Yep, that sounds precise

src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -174,3 +174,59 @@ export const getUserMinStake = (
...
+        name: 'MotionStaked',
+        messageKey,
+        initiatorAddress: staker,
+        vote: vote.toString(),
+        amount: amount.toString(),
+      },
+    );
+
+    if (vote.eq(1) && requiredStake.eq(motionStakes.raw.yay)) {
+      const messageName = !updatedMessages.find(({ name }) => name === 'ObjectionRaised') ? 'MotionFullyStaked' : 'MotionFullyStakedAfterObjection';
```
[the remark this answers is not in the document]
Julianahlk: @willm30 I added the boolean flag as requested `hasObjection`. Can you double check this is what you intended? Thank you.

src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -182,3 +182,71 @@ export const getUserMinStake = (
...
+      {
+        name: messageName,
+        messageKey: `${messageKey}_${messageName}`,
+        initiatorAddress: staker,
+      },
+    );
+  }
+
+  if (vote.eq(MotionVote.NAY) && requiredStake.eq(motionStakes.raw.nay) && BigNumber.from(motionStakes.raw.nay).gt(motionStakes.raw.yay)) {
+    updatedMessages.push(
```
willm30, line 242: How come we have this check here also?:

`BigNumber.from(motionStakes.raw.nay).gt(motionStakes.raw.yay))`

Aren't these are sufficient to determine if it's fully staked?

``` js
vote.eq(MotionVote.NAY) && requiredStake.eq(motionStakes.raw.nay)
```

src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -182,3 +182,71 @@ export const getUserMinStake = (
     .div(BigNumber.from(10).pow(36))
     .toString();
 };
+
+interface Props {
+  motionData: MotionData,
```
willm30, line 187: Commas should be semi-colons

willm30: Are you using prettier? It catches things like this automatically

src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -182,3 +182,71 @@ export const getUserMinStake = (
...
+  staker,
+  amount,
+}: Props): MotionMessage[] => {
+  const updatedMessages = [...messages];
+  const isFirstObjection = vote.eq(MotionVote.NAY) && !motionData.hasObjection;
+  if (isFirstObjection) {
+    motionData.hasObjection = true;
+    updatedMessages.push(
+      {
+        name: 'ObjectionRaised',
```
willm30, line 213: These strings should come from an enum, like in the fe

src/handlers/motions/motionStaked/helpers.ts
willm30: In any case, a comment in the code would be useful here

Julianahlk: @willm30 I added this based on your comment in the cdapp PR that we shouldn't show ObjectionFullyStaked if the motion is already fully staked in favor. `BigNumber.from(motionStakes.raw.nay).gt(motionStakes.raw.yay)` is the same logic in the dapp to handle this situation. The only way the nay vote could be greater than the yay vote is if the motion was not already fully staked for the yay vote. I can add a comment to explain.

src/handlers/motions/motionStaked/helpers.ts
Julianahlk: @willm30 I fixed the formatting issues. I needed to create a .prettierrc file to get prettier to run correctly, how are you using prettier in the block ingestor?

src/handlers/motions/motionStaked/helpers.ts
Julianahlk: great idea, thanks

src/handlers/motions/motionStaked/helpers.ts
willm30: I have it installed in vsc

Julianahlk: I had it installed in vsc but not sure why it wasn't working. Thanks again for pointing it out.

PR #46, opened 2023-04-06
Add voting support (motions)

willm30:
This PR adds support for the `MotionVoteSubmitted` event (i.e. motion voting).

Test with: https://github.com/JoinColony/colonyCDapp/pull/380

### DIFF
--- a/src/handlers/motions/motionVoteSubmitted/helpers.ts
+++ b/src/handlers/motions/motionVoteSubmitted/helpers.ts
@@ -1,0 +1,35 @@
+import { BigNumber } from 'ethers';
+import { VoterRecord } from '~types';
+
+export const getUpdatedVoterRecord = (
+  voterRecord: VoterRecord[],
+  voterAddress: string,
+): VoterRecord[] => {
+  const currentVoterRecord = voterRecord.find(
+    ({ address }) => address === voterAddress,
+  );
+
+  if (!currentVoterRecord) {
+    return [
+      ...voterRecord,
+      {
+        address: voterAddress,
+        voteCount: '1',
+      },
+    ];
+  }
+
+  return voterRecord.map((record) => {
+    const { address } = record;
+    if (address === voterAddress) {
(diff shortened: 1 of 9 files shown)


rdig (approved):
All seems good, and as far as I can tell it also supports the user changing their vote, prior to the reveal period

src/handlers/motions/motionVoteSubmitted/helpers.ts
```diff
@@ -0,0 +1,35 @@
...
+    ];
+  }
+
+  return voterRecord.map((record) => {
+    const { address } = record;
+    if (address === voterAddress) {
+      const { voteCount } = currentVoterRecord;
+      const updatedRecord: VoterRecord = {
+        ...currentVoterRecord,
+        voteCount: BigNumber.from(voteCount).add(1).toString(),
```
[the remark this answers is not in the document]
willm30: It's a counter for how many times a specific user voted. I'm using it to detect in the front end if a user changed their vote.

src/handlers/motions/motionVoteSubmitted/index.ts
```diff
@@ -0,0 +1,40 @@
...
+  const votingClient = await getVotingClient(colonyAddress);
+  const { chainId } = await votingClient.provider.getNetwork();
+  const motionDatabaseId = getMotionDatabaseId(
+    chainId,
+    votingClient.address,
+    motionId,
+  );
+  const votedMotion = await getMotionFromDB(colonyAddress, motionDatabaseId);
+
+  if (votedMotion) {
```
jakubcolony, line 25: Do we want to log when that's not found?

jakubcolony (approved):
Nice work! 👍

src/handlers/motions/motionVoteSubmitted/index.ts
willm30: @jakubcolony It gets logged inside `getMotionFromDB`

PR #48, opened 2023-04-10
Add reveal support

willm30:
This PR adds support for the Reveal widget. Test with https://github.com/JoinColony/colonyCDapp/pull/385

### DIFF
--- a/src/handlers/motions/motionVoteRevealed/index.ts
+++ b/src/handlers/motions/motionVoteRevealed/index.ts
@@ -1,0 +1,54 @@
+import { ContractEvent, VoterRecord } from '~types';
+import { getVotingClient, verbose } from '~utils';
+import {
+  getMotionDatabaseId,
+  getMotionFromDB,
+  updateMotionInDB,
+} from '../helpers';
+
+export default async (event: ContractEvent): Promise<void> => {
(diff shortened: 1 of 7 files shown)


rdig (approved):
Yup, seems to work as good

src/handlers/motions/motionVoteRevealed/index.ts
```diff
@@ -0,0 +1,54 @@
...
+  const votingClient = await getVotingClient(colonyAddress);
+  const { chainId } = await votingClient.provider.getNetwork();
+  const motionDatabaseId = getMotionDatabaseId(
+    chainId,
+    votingClient.address,
+    motionId,
+  );
+  const revealedMotion = await getMotionFromDB(colonyAddress, motionDatabaseId);
+
+  if (revealedMotion) {
```
jakubcolony, line 24: As in the other one, should we log if not found?

jakubcolony (approved):
All good here! 👍

src/handlers/motions/motionVoteRevealed/index.ts
willm30: Gets logged inside getMotionFromDB

PR #51, opened 2023-04-12
Add support for Voting Progress

willm30:
This PR adds support for the VotingProgress component: https://github.com/JoinColony/colonyCDapp/pull/401

Specifically: 

1. Stores voting rep params in db on initialisiation
2. Adds skillRep and repSubmitted to motion data

### DIFF
--- a/src/handlers/extensions/helpers.ts
+++ b/src/handlers/extensions/helpers.ts
@@ -1,0 +1,50 @@
+import { mutate } from '~amplifyClient';
+import { ExtensionParams } from '~types';
+import { getVotingClient } from '~utils';
+
+const getVotingReputationParams = async (
+  colonyAddress: string,
+): Promise<ExtensionParams> => {
+  const votingClient = await getVotingClient(colonyAddress);
+  const totalStakeFraction = (
+    await votingClient.getTotalStakeFraction()
+  ).toString();
+  const voterRewardFraction = (
+    await votingClient.getVoterRewardFraction()
+  ).toString();
+  const userMinStakeFraction = (
+    await votingClient.getUserMinStakeFraction()
+  ).toString();
+  const maxVoteFraction = (await votingClient.getMaxVoteFraction()).toString();
+  const stakePeriod = (await votingClient.getStakePeriod()).toString();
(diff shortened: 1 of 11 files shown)


src/handlers/extensions/helpers.ts
```diff
@@ -0,0 +1,50 @@
...
+};
+
+export const addVotingReputationParamsToDB = async (
+  extensionAddress: string,
+  colonyAddress: string,
+): Promise<void> => {
+  await mutate('updateColonyExtensionByAddress', {
+    input: {
+      id: extensionAddress,
+      params: await getVotingReputationParams(colonyAddress),
```
rdig, line 47: I'm not a fan of this pattern personally. I'd have this assigned to a variable then pass that along to the mutation input.

rdig (approved):
Fine by me. Left a comment, but it's more of a nitpick rather than something that needs to be addressed

src/trackExtensions.ts
```diff
@@ -1,5 +1,7 @@
 import { Extension, getExtensionHash, getLogs } from '@colony/colony-js';
 
+import { addVotingReputationParamsToDB } from '~handlers/extensions';
```
jakubcolony, line 3: Nitpick, but having to import it from `~handlers` suggests to me it could live in `~utils` instead

jakubcolony (approved):
All good, one thing that relates more to the CDapp part of this (that I was too slow to review) is that we could use a union type for extension params instead, just to reduce the nesting (and since there can only be one type of params for an extension), e.g.:

```gql
type VotingReputationParams {
  ...
}

union ExtensionParams = VotingReputationParams | OtherParamsInTheFuture;

type ColonyExtension {
  ...
  params: ExtensionParams
}
```

jakubcolony:
@willm30 Was that not the case only for models?

jakubcolony:
I got the schema to compile with a union. Haven't tried actually querying it though

willm30:
@jakubcolony Not certain but even on non-models, we'd still need to write some logic to tell Amplify which type it resolves to given the surrounding data (in this case, the extension hash).

PR #52, opened 2023-04-13
Fix staking percentage calculation when staking min

willm30:
This PR ensures that the staking percentage is never calculated as zero, since it's impossible for a user to stake 0% of a motion. 

This might happen due to rounding when staking the minimum stake. In this case, we round up to 1.


rdig (approved):
:+1:

jakubcolony (approved):
👍 👍 👍

PR #53, opened 2023-04-14
Add actions list support

willm30:
This PR adds a flag to actions and motions indicating whether they should appear in the actions list. Test with https://github.com/JoinColony/colonyCDapp/pull/410

It's:

- true for actions that were not initiated by a voting rep extn
- true for motions if total stake >= 10 %
- else false

### DIFF
--- a/src/amplifyClient.ts
+++ b/src/amplifyClient.ts
@@ -216,6 +216,23 @@ export const queries = {
       }
     }
   `,
+  getVotingRepInstallations: /* GraphQL */ `
+    query GetVotingRepInstallations(
+      $votingRepHash: String!
+      $colonyAddress: ID!
+    ) {
+      listColonyExtensions(
+        filter: {
+          hash: { eq: $votingRepHash }
+          colonyId: { eq: $colonyAddress }
+        }
+      ) {
+        items {
+          id
+        }
+      }
+    }
+  `,
 };
 
 export default (): void => {
--- a/src/amplifyClient.ts
+++ b/src/amplifyClient.ts
@@ -221,11 +221,9 @@ export const queries = {
       $votingRepHash: String!
       $colonyAddress: ID!
     ) {
-      listColonyExtensions(
-        filter: {
-          hash: { eq: $votingRepHash }
-          colonyId: { eq: $colonyAddress }
-        }
+      getExtensionByColonyAndHash(
+        colonyId: $colonyAddress
+        hash: { eq: $votingRepHash }
       ) {
         items {
           id
--- a/src/handlers/motions/motionStaked/motionStaked.ts
+++ b/src/handlers/motions/motionStaked/motionStaked.ts
@@ -24,7 +24,9 @@ export default async (event: ContractEvent): Promise<void> => {
   const requiredStake = getRequiredStake(skillRep, totalStakeFraction);
   const motionStakes = getMotionStakes(requiredStake, stakes);
   const remainingStakes = getRemainingStakes(requiredStake, stakes);
-
+  const showInActionsList =
+    Number(motionStakes.percentage.yay) + Number(motionStakes.percentage.nay) >=
+    10;
   const { chainId } = await votingClient.provider.getNetwork();
   const motionDatabaseId = getMotionDatabaseId(
     chainId,
@@ -48,12 +50,16 @@ export default async (event: ContractEvent): Promise<void> => {
       requiredStake,
     );
 
-    await updateMotionInDB(id, {
-      ...motionData,
-      usersStakes: updatedUserStakes,
-      motionStakes,
-      remainingStakes,
-    });
+    await updateMotionInDB(
+      id,
+      {
+        ...motionData,
+        usersStakes: updatedUserStakes,
+        motionStakes,
+        remainingStakes,
+      },
+      showInActionsList,
+    );
 
     verbose(
       `User: ${staker} staked motion ${motionId.toString()} by ${amount.toString()} on side ${getMotionSide(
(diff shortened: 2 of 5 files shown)


src/utils/actions/writeAction.ts
```diff
@@ -1,24 +1,65 @@
-import { mutate } from '~amplifyClient';
-import { ContractEvent } from '~types';
+import { Extension, getExtensionHash } from '@colony/colony-js';
+import { mutate, query } from '~amplifyClient';
+import { ColonyActionType, ContractEvent } from '~types';
 import { verbose } from '~utils';
 
+interface ActionFields extends Record<string, any> {
+  initiatorAddress: string;
```
willm30, line 7: Now that the initiatorAddress is needed to determine whether to show in the actions list, making it explicit here for type safety

src/handlers/motions/motionStaked/motionStaked.ts
```diff
@@ -24,7 +24,9 @@ export default async (event: ContractEvent): Promise<void> => {
   const requiredStake = getRequiredStake(skillRep, totalStakeFraction);
   const motionStakes = getMotionStakes(requiredStake, stakes);
   const remainingStakes = getRemainingStakes(requiredStake, stakes);
-
+  const showInActionsList =
```
willm30, line 27: We put a motion in the actions list if it has been staked above 10%

src/utils/actions/writeAction.ts
```diff
@@ -1,24 +1,65 @@
...
+      {
+        votingRepHash,
+        colonyAddress,
+      },
+    )) ?? {};
+
+  return items;
+};
+
+const showActionInActionsList = async (
```
willm30, line 53: We put an action in the actions list if it wasn't created by a motion. This is to avoid showing a duplicate once the motion has been finalized.

src/amplifyClient.ts
```diff
@@ -216,6 +216,23 @@ export const queries = {
...
+          hash: { eq: $votingRepHash }
+          colonyId: { eq: $colonyAddress }
+        }
+      ) {
+        items {
+          id
+        }
+      }
+    }
+  `,
```
jakubcolony, line 235: That might break relatively soon. By default, it'll only get the first 100 of all the extensions ever installed and filter those for the right hash and colony address.
Have a look at `getExtensionByColonyAndHash` query, despite its slightly imprecise name it should return all installations for a given colony and hash using indexes on these fields instead of filtering

src/handlers/motions/motionStaked/motionStaked.ts
rdig: So the way I see it, they need to be in the database iregardless of the percent of stake they have, and have them filtered by that percent, only when querying via graphQL

This is because of couple of reasons:
- we might change that percent value (which means we'll not be able to show ongoing / historic motions since as far as our db is concerned they won't exist)
- there was talk at some point that we might wave that limitation for certain users (more reputation, or some permissions)

src/utils/actions/writeAction.ts
jakubcolony: Could we potentially check for that when the action is handled instead? And not write it to the DB at all? I don't know if we need that info, but it'd simplify all the surrounding logic

src/utils/actions/writeAction.ts
```diff
@@ -1,24 +1,65 @@
...
+  colonyAddress: string,
+  initiatorAddress: string,
+): Promise<boolean> => {
+  const votingRepIds = await getVotingReputationInstallations(colonyAddress);
```
rdig, line 57: I wanted to complain about this, and have you fetch the extension from the chain directly (since you can just get the address directly, and it'll be faster than graphql) but I've come to understand your approach is better.

If you use my approach, you can only filter using the currently installed extension address (if it exists), all other actions, for earlier uninstalled extensions would show up at that point.

So your approach is better in this case :)

rdig (approved):
Good, however the motions need to be stored in the db, no matter the stake, and just filter them when fetching them on the client.

src/handlers/motions/motionStaked/motionStaked.ts
jakubcolony: It's not ideal that block-ingestor knows of that, but it's also hard to think about a good alternative. I feel like it'd be better to filter for this condition in the CDapp (we do it regardless for the `showInActionsList` field, and it's a UI detail that I think belongs in the data presentation) but it's not a single field that you can filter on in your query...

Maybe if we can stop writing motion-created actions in the DB (see my later question) this will be irrelevant?

jakubcolony (changes requested):
Looking good, but please have a look at the `GetVotingRepInstallations` query

jakubcolony:
> Could we call to discuss your other comments, and if necessary, address in a follow up issue?

Sure, feel free to ping me or put something in my calendar

>I also feel it's consistent with the general approach of keeping logic off the client where possible.

I'm under impression the approach is to keep the processing of the data and facilitate caching of the blockchain events/transactions off the client, more than the logic. I see block-ingestor as one service in microservices architecture that should ideally be as independent and agnostic as possible, ie. not care/know too much about how the data is presented, but only process and store it. To me, `showInActionsList` feels like a UI rule that should be kept with the UI. Especially that doing it here doesn't give us any performance advantage. 

But it doesn't solve the bigger problem: Getting the 10 filtered actions, not 10 actions that you still need to filter. 

>If it proves challenging to implement with amplify, then I think there's a case for moving the logic to the client (I actually implemented this initially

But even moving the filtering to the client doesn't ensure you'll end up with 10 actions _after_ filtering, no? Or did you managed to solve this somehow?

If you can't have a key on a boolean, then there might be a case for making the property on which the filtering actually depends (the sum of yay and nay percentages I think?) a filterable key, we should be able to just store it as a number (since it'll be between 0 and 100). And add the `gt`/`lt` filter as a property in the CDapp's query.

jakubcolony (approved):
Approving following our call earlier, if you want me to look at any further changes just ping me

PR #57, opened 2023-04-18
Add payment motion support

willm30:
This pr adds support for the payment motion. Test with: https://github.com/JoinColony/colonyCDapp/pull/430/

### DIFF
--- a/src/handlers/motions/motionCreated/helpers.ts
+++ b/src/handlers/motions/motionCreated/helpers.ts
@@ -1,8 +1,7 @@
-import { AnyColonyClient, Extension } from '@colony/colony-js';
 import { BigNumber } from 'ethers';
 import { TransactionDescription } from 'ethers/lib/utils';
 
-import { MotionData, ContractEvent } from '~types';
+import { ContractEvent, MotionData, NetworkClients } from '~types';
 import { getVotingClient, verbose } from '~utils';
 
 import { mutate } from '~amplifyClient';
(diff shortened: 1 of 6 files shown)


src/handlers/motions/motionCreated/helpers.ts
```diff
@@ -14,22 +13,25 @@ import {
...
   const motion = await votingClient.getMotion(motionId);
 
-  try {
-    return colonyClient.interface.parseTransaction({
-      data: motion.action,
-    });
-  } catch {
-    verbose(`Unable to parse ${motion.action}`);
-    return undefined;
+  for (const client of clients) {
```
willm30, line 22: We don't know which client is going to be needed to parse the transaction just from the event data... right?

On this assumption, I'm looping through the set of possible clients and returning the first time we don't get an error.

rdig: Correct, and this is the right approach.

We have done this in the past, both for motions and both for normal events

rdig (approved):
All good

ArmandoGraterol (approved):
Good one!

src/handlers/motions/motionCreated/helpers.ts
```diff
@@ -14,22 +13,25 @@ import {
 
 export const getParsedActionFromMotion = async (
   motionId: string,
-  colonyClient: AnyColonyClient,
+  colonyAddress: string,
+  clients: NetworkClients[],
```
jakubcolony, line 17: Would it make sense to restrict this type to only the clients that can parse actions? I think this currently also includes network and token clients

jakubcolony (approved):
Looking good 👍

PR #60, opened 2023-04-24
Handle Permissions Events

rdig:
This PR adds in logic to handle listening to permissions role events, create database entries for them, create historic entries, and finally create permissions actions.

To test this head on over to the CDapp and test the JoinColony/colonyCDapp#271

Some of the logic required gets pretty harry, and so I've left a bunch of comments all over the place. Please make sure to give them a read, to understand the flow of data through the ingestor.


src/handlers/actions/managePermissions.ts
```diff
@@ -0,0 +1,127 @@
...
+export default async (event: ContractEvent): Promise<void> => {
+  const { args, contractAddress, blockNumber, transactionHash } = event;
+  const { agent, user: targetAddress, domainId } = args;
+
+  const id = getColonyRolesDatabaseId(contractAddress, domainId.toString(), targetAddress);
+  const domainDatabaseId = getDomainDatabaseId(contractAddress, domainId.toString());
+
+  /*
+   * @TODO
+   * - create action list entry
```
jakubcolony, line 23: Doesn't this happen in line 80 already? Or is that about something else?

src/handlers/actions/managePermissions.ts
```diff
@@ -0,0 +1,127 @@
...
+            transactionHash,
+            logIndex,
+          }) => ({
+            id: `${transactionHash}_${logIndex}`,
+            type: name,
+            role,
+            setTo,
+          })),
+        ),
+      });
```
jakubcolony, line 101: We're doing a similar mapping twice so maybe a util would come handy?

src/utils/permissions.ts
```diff
@@ -0,0 +1,321 @@
...
+export const createColonyFounderInitialRoleEntry = async (event: ContractEvent): Promise<void> => {
+  const { name, transactionHash, args } = event;
+  const { colonyAddress } = args;
+
+  const ColonyAddedEventName = ContractEventsSignatures.ColonyAdded.slice(0, ContractEventsSignatures.ColonyAdded.indexOf('('));
+  const ColonyRoleSetEventName = ContractEventsSignatures.ColonyRoleSet.slice(0, ContractEventsSignatures.ColonyRoleSet.indexOf('('));
+
+  if (name !== ColonyAddedEventName) {
+    throw new Error('The event passed in is not the "ColonyAdded" event. We can\'t determine the colony\'s founder otherwise');
+  }
```
jakubcolony, line 252: Could you not compare the signatures instead?

jakubcolony (approved):
As much as I can tell, it looks good 👍

src/eventProcessor.ts
```diff
@@ -158,6 +159,21 @@ export default async (event: ContractEvent): Promise<void> => {
       return;
     }
 
+    case ContractEventsSignatures.ColonyRoleSet: {
```
willm30, line 162: This could read: 
``` js
case ContractEventsSignatures.ColonyRoleSet: 
case ContractEventsSignatures.ColonyRoleSet_OLD:
case ContractEventsSignatures.RecoveryRoleSet: {
 await handleManagePermissionsAction(event);
      return;
}
```

src/handlers/actions/managePermissions.ts
```diff
@@ -0,0 +1,127 @@
...
+    ...existingRoles
+  } = (await query('getColonyRole', { id })) || {};
+
+  /*
+   * update the entry
+   */
+  if (existingColonyRoleId) {
+    /*
+     * Only update the entry if the event we're processing is older than the latest
+     * permissions we stored in the database
```
willm30, line 38: Nitpick: "older" is ambiguous here, we mean "more recent", right?

src/handlers/actions/managePermissions.ts
```diff
@@ -0,0 +1,127 @@
...
+     *
+     * Would be ideal if we could tell if the previous mutation executed, because as it stands,
+     * until it finishes, this logic check can't really tell and will try to execute it again
+     *
+     * (it doesn't break anything, as the GraphQL server will just discard it with an error,
+     * but it just adds more travel time which is wasted)
+     */
+    if (blockNumber > parseInt(existingColonyRoleLatestBlock, 10)) {
```
willm30, line 46: `existingColonyRoleLatestBlock` is stored as an Int, so why use `parseInt`?

src/utils/permissions.ts
```diff
@@ -0,0 +1,321 @@
+/* eslint-disable @typescript-eslint/naming-convention */
```
willm30, line 1: Nitpick: this doesn't need to be disabled for the whole file, just lines 117-122

src/utils/permissions.ts
```diff
@@ -0,0 +1,321 @@
...
+  userAddress: string,
+  blockNumber: number,
+): string => `${colonyAddress}_${nativeDomainId}_${userAddress}_${blockNumber}_roles`;
+
+export const getAllRoleEventsFromTransaction = async (
+  transactionHash: string,
+  colonyAddress: string,
+): Promise<ContractEvent[]> => {
+  const colonyRoleSetEventName = ContractEventsSignatures.ColonyRoleSet.slice(0, ContractEventsSignatures.ColonyRoleSet.indexOf('('));
+  const colonyRecoveryRoleSetName = ContractEventsSignatures.RecoveryRoleSet.slice(0, ContractEventsSignatures.RecoveryRoleSet.indexOf('('));
```
willm30, line 43: "RecoveryRoleSet"?

src/utils/permissions.ts
```diff
@@ -0,0 +1,321 @@
...
+
+  const colonyClient = await getCachedColonyClient(colonyAddress);
+  const transactionReceipt = await colonyClient.provider.getTransactionReceipt(transactionHash);
+
+  const events = await Promise.all(
+    transactionReceipt.logs.map(log => mapLogToContractEvent(log, colonyClient.interface)),
+  );
+
+  const filteredEvents = events.filter(event => {
```
willm30, line 52: Any chance you could run prettier on this pr so we keep style consistent? I know there's an issue for that to be handled automatically in this repo, but for the meantime?

src/handlers/actions/managePermissions.ts
```diff
@@ -0,0 +1,127 @@
...
+      const allRoleEventsUpdates = await getAllRoleEventsFromTransaction(transactionHash, contractAddress);
```
willm30, line 47: Maybe leave a comment here as this deviates from the general pattern of having one handler per event. I was scratching my head as to why we even need to do this, before realising it's because we have one handler here for multiple events.

src/utils/permissions.ts
willm30: This filter fn could be more succinct:

``` js
events.filter((event) => event?.name === colonyRoleSetEventName || event?.name === colonyRecoveryRoleSetName)
```

src/utils/permissions.ts
```diff
@@ -0,0 +1,321 @@
...
+    if (
+      !event ||
+      !(event.name === colonyRoleSetEventName || event.name === colonyRecoveryRoleSetName)
+    ) {
+      return false;
+    }
+    return true;
```
willm30, line 59: This filter fn could be more succinct:

``` js
events.filter((event) => event?.name === colonyRoleSetEventName || event?.name === colonyRecoveryRoleSetName)
```

src/handlers/actions/managePermissions.ts
willm30: If it sounds like I'm off or would be easier to explain on a call, I'd happy  jump on one to get a better understanding of how this all fits together (and would probably be beneficial for everyone also)

src/utils/permissions.ts
```diff
@@ -0,0 +1,321 @@
...
+     * It's role is to format the object key/value pairs for roles (eg: `{ role_1: true }`).
+     *
+     * Since we have a separate event for `RecoveryRoleSet` which has a different format than the rest, check for it, and convert the
+     * role to `0` if we find it.
+     *
+     * Second part (after the || operator check) deals with how we display the "off" side of the role. To keep database consistency, for operational
+     * roles (ie: the ones we use to check against), we check the `false` value (which signifies the permission was taken away) to `null` as that is
+     * the default state of the GraphQL entry. For Actions however, we need to set it to `false` as to distiguish itself, and for us to know what
+     * to display in the Action's text. For this we use `setToNull`, which, when set (on by default), changes the "off" state to null, and when it's
+     * not being set, changes the "off" state to false
```
willm30, line 88: Hard to follow. I think this has something to do with distinguishing between whether a role was set to "off" versus whether it has not been set yet at all?

src/utils/permissions.ts
```diff
@@ -0,0 +1,321 @@
...
+   * Typecasting since apparently TS doesn't realize we are actually filtering
+   * to ensure that the Array only contains proper events
+   */
+  return filteredEvents as ContractEvent[];
+};
+
+export const getRolesMapFromEvents = (roleEvents: ContractEvent[], setToNull: boolean = true): Record<string, boolean | null> => {
+  let roleMap = {};
+
+  roleEvents.map(({
```
willm30, line 72: This should use `.reduce`, right?

willm30 (commented):
My main question is what is the advantage of processing all RoleSet events in a given block at once, instead of processing them individually? I feel like the batch processing approach adds some complexity compared to making db mutatations on a per event basis (assuming we can link ColonyRoleSet events to their respective ColonyRole and ColonyAction db entries)

src/utils/permissions.ts
rdig: For colony roles / historic roles we only care about if the role is "on" (set to `true`), while everything else that's "off" is set to the `null` which is the default for graphQL (i didn't want to mix both `false` and `null` as to mean "off")

For action roles however, I explicitly need to know if a role was "turned on" (set to `true`) or if it was "turned off" (set to `false`), and I only care about the ones that were changed, not the ones that weren't touched (so that we can generate the action title. ie: role X was added, role Y was removed)

src/utils/permissions.ts
jakubcolony: Ah fair enough. Then how about a util to get the event name from its signature?

rdig:
As with it's CDapp counterpart, this has been rebased, tested, and it's ready for re-review/merge.

As with that one, the more time consuming changes that were pointed out were extracted into a separate issue: #79

This was done in the interest of time, as to get this merged faster, and push forward towards production deployment.

jakubcolony (approved):
Added a couple comments

willm30 (approved):
👍

jakubcolony (approved):
👍

PR #67, opened 2023-05-01
Add formatting & linting

jakubcolony:
This PR adds a few components to the formatting/linting setup of block-ingestor to make it more consistent with the CDapp:
- Prettier config (with relevant ESLint plugin)
- lint-staged config responsible for linting and formatting 
- husky hooks to run lint-staged on commit and typecheck on push

## Testing
- Run `npm i`
- Try to commit a file with eslint error (it should either be fixed automatically or committing should fail)
- Try to commit a file with formatting issue (it should be formatted automatically)
- Try pushing to remote when there's a TypeScript error present (doesn't have to be committed; the push should fail)


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged
```
willm30, line 4: new line here?

ArmandoGraterol (approved):
Works for me 👌

Julianahlk (approved):
Nice! Glad to have this set up in the block-ingestor now. 👍

PR #68, opened 2023-05-01
Add GraphQL codegen

jakubcolony:
This PR adds codegen script that generates GraphQL typings based on the queries, mutations and fragments.

The new workflow for querying Amplify API:
1. In block-ingestor, run `npm run codegen`. This will start codegen in watch mode.
2. Create/edit a .graphql file in `queries`, `mutations`, or `fragments` folder
3. The codegen will pick up those changes and generate `graphql/generated.ts` file

Note that whenever the schema changes, you will need to restart the codegen script. We have a similar issue in the CDapp, and I don't think there's a simple solution to that (the only one I can think of is polling the schema and manually comparing for changes, which is doable, but time-consuming). Because of that I opted for a manually-run codegen script rather than running it alongside `npm run dev`.

The generated types file will contain types for all GraphQL Object Types, Inputs, Enums, etc. present in the schema. It will also contain the types of client-side defined operations (queries, mutations) and fragments. For each operation, a `DocumentNode` will be generated which will contain the actual operation definition.

I modified the existing `query` and `mutations` functions accordingly so that they can be used in the following way:
```ts
import {
  SetCurrentVersionMutation,
  SetCurrentVersionMutationVariables,
  SetCurrentVersionDocument,
} from '~graphql';

await mutate<SetCurrentVersionMutation, SetCurrentVersionMutationVariables>(
  SetCurrentVersionDocument,
  {
    // the variables go here
  },
);
```

## Testing
Check that all block-ingestor functionality still works. Run the script, edit the existing queries or add new ones, see that you get the correctly generated types.


willm30 (approved):
LGTM, nice one Jakub!

Julianahlk (commented):
I had a question about how to test this, is this expected?

If I directly do `npm run dev` it works fine but
If I do `npm run codegen` then `npm run dev` I get an updated generated.ts file and the error below (this is without making any changes to the project):

`[ERROR] 21:11:50 ⨯ Unable to compile TypeScript:
src/handlers/actions/mintTokens.ts(11,52): error TS2345: Argument of type '{ type: ColonyActionType.MintTokens; initiatorAddress: any; recipientAddress: any; amount: any; tokenAddress: string; }' is not assignable to parameter of type 'Omit<CreateColonyActionInput, "blockNumber" | "colonyId" | "colonyActionsId">'.
  Property 'showInActionsList' is missing in type '{ type: ColonyActionType.MintTokens; initiatorAddress: any; recipientAddress: any; amount: any; tokenAddress: string; }' but required in type 'Omit<CreateColonyActionInput, "blockNumber" | "colonyId" | "colonyActionsId">'.`

jakubcolony:
@Julianahlk Which CDapp branch have you got checked out?

Julianahlk (approved):
Looking good now. Confirmed I am running the Cdapp from master to test.

PR #69, opened 2023-05-02
Fix rounding up to one for wrong stake side (motions)

willm30:
This PR is a small fix to the staking calculation. I previously changed the logic so that if the stake percentage was 0, we'd show 1, since it'd only be 0 as a result of BigNumber rounding down (you cannot stake 0% of a motion). However, I failed to discriminate by vote side, meaning it currently affects both sides (which it shouldn't, since you can only stake one side at a time). 

This PR fixes that, such that staking one side doesn't cause the other side to be rounded up in the event its stake percentage is 0. 

You can test on the main cdapp motions branch. Check out and run block ingestor separately. Object a motion, and the yay side should remain at 0% and vice versa.

### DIFF
--- a/src/handlers/motions/motionStaked/helpers.ts
+++ b/src/handlers/motions/motionStaked/helpers.ts
@@ -18,16 +18,24 @@ export const getRequiredStake = (
 export const getMotionStakes = (
   requiredStake: BigNumber,
   [totalNayStakesRaw, totalYayStakesRaw]: [BigNumber, BigNumber],
+  vote: BigNumber,
 ): MotionStakes => {
-  const totalYayStakesPercentage = getStakePercentage(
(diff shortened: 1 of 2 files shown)


Julianahlk (commented):
Just wanted to clarify the expected behavior. In this example, all I did was object and the Motion is now staked at 1%. Is this what you would expect?

Screenshot 2023-05-02 at 9 48 15 PM

willm30:
@Julianahlk That is the behaviour this pr should resolve. Can I double check you were testing with this hash of the block ingestor?

Julianahlk (approved):
I tested again today and double checked I was running the correct block ingestor and its working as expected now. 👍 

Screenshot 2023-05-03 at 8 38 00 PM

ArmandoGraterol (approved):
Working as expected 👍

PR #71, opened 2023-05-04
Track extensions, colonies and actions from the latest processed block

jakubcolony:
This optimisation allows block-ingestor to pick up right where it left off before restart, rather than having to go from block 1. It uses `node-persist` library for easier persistent storage (which still uses JSON files, but has a nicer interface)

### Testing
I imagine this approach will only get tested property in the QA environment, but for now:
1. Run the ingestor on this branch separately
2. Run the `temp-create-data` script in CDapp
3. Restart the ingestor. On `master`, you'd see a bunch of failed mutations related to extensions and actions. On this branch, there shouldn't be more than 1-2 failed mutations (I suspect these are due to ganache/block-ingestor replaying some of the transactions)

TODO: 
- [x] Refactor stats to use `node-persist`
- [x] Track extensions from latest processed block
- [x] Track actions from latest processed block

### DIFF
--- a/stats/446501053769c06c565094b26d26e8ef
+++ b/stats/446501053769c06c565094b26d26e8ef
@@ -1,1 +1,0 @@
-{"key":"stats","value":{"trackedColonies":5,"latestBlock":1039}}
--- a/stats/446501053769c06c565094b26d26e8ef
+++ b/stats/446501053769c06c565094b26d26e8ef
@@ -1,0 +1,1 @@
+{"key":"stats","value":{"trackedColonies":5,"latestBlock":1039}}
--- a/.gitignore
+++ b/.gitignore
@@ -103,4 +103,4 @@ dist
 # TernJS port file
 .tern-port
 
-run
+stats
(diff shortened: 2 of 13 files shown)


willm30:
@jakubcolony Is this ready for review?

> (I suspect these are due to ganache/block-ingestor replaying some of the transactions)

I believe this is due to ethers v5 processing all events in the current block, even if they've already been processed. There's a solution suggested here: https://github.com/ethers-io/ethers.js/issues/2310

willm30 (approved):
LGTM, nice job!

stats/446501053769c06c565094b26d26e8ef
willm30, line 1: Should we be committing this file? It will always be changing when used in development.

jakubcolony: Most likely not. The only thing to consider is that it persist between different builds of block-ingestor. Let me check with Raul on that

ArmandoGraterol (approved):
Solid work 👍

PR #73, opened 2023-05-12
Add missing `fromDomainId` to `mintTokens` handler

ArmandoGraterol:
> The action list item for the mint tokens action is not displaying the team in the meta info, which should be "Root”:
FireShot Capture 039 - Colony CDapp - qa-cdapp colony io

Found while testing on QA.

To test this, just replace the commit hash in `master`, create a mint tokens action, and check that the meta info of the action list item includes the `Root` domain:

FireShot Capture 053 - Colony CDapp - localhost

### DIFF
--- a/src/handlers/actions/mintTokens.ts
+++ b/src/handlers/actions/mintTokens.ts
@@ -1,6 +1,12 @@
-import { ColonyActionType } from '~graphql';
+import { Id } from '@colony/colony-js';
+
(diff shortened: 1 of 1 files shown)


jakubcolony (approved):
Nice catch & fix! 👍

Julianahlk (approved):
Nice fix 👍

PR #74, opened 2023-05-16
Wire Manage Tokens / Edit Colony motion

willm30:
This PR adds support for the manage tokens motion. Test with https://github.com/JoinColony/colonyCDapp/pull/553.

It works the same as Domains.

In the editColonyMotion saga, we create a "pending" colonyMetadata object that represents the intended state change.

When a successful motion is finalized, we update the actual colony metadata with the changes, and update the colony tokens if needed. 

Note that I've implented this so that, in theory, we won't be overwriting metadata state changes that occurred after the motion was created (as currently happens with the domain metadata). I have only been able to test that this is true for adding / removing tokens as other colonyEditMotion functionality is yet to be wired (this should also be tested then).

### DIFF
--- a/src/handlers/motions/motionCreated/handlers/editColony.ts
+++ b/src/handlers/motions/motionCreated/handlers/editColony.ts
@@ -3,19 +3,17 @@ import { ContractEvent, motionNameMapping } from '~types';
 import { getPendingMetadataDatabaseId } from '~utils';
 import { createMotionInDB } from '../helpers';
 
-export const handleManageDomainMotion = async (
+export const handleEditColonyMotion = async (
   event: ContractEvent,
-  parsedAction: TransactionDescription,
+  { name }: TransactionDescription,
 ): Promise<void> => {
   const { transactionHash, contractAddress: colonyAddress } = event;
-  const { name } = parsedAction;
-  const pendingDomainMetadataId = getPendingMetadataDatabaseId(
+  const pendingColonyMetadataId = getPendingMetadataDatabaseId(
     colonyAddress,
     transactionHash,
   );
-
   await createMotionInDB(event, {
     type: motionNameMapping[name],
-    pendingDomainMetadataId,
+    pendingColonyMetadataId,
   });
 };
--- a/src/handlers/motions/motionFinalized/helpers.ts
+++ b/src/handlers/motions/motionFinalized/helpers.ts
@@ -1,10 +1,23 @@
 import { BigNumber } from 'ethers';
 import { TransactionDescription } from 'ethers/lib/utils';
 
-import { ColonyOperations, MotionQuery, MotionVote, StakerReward } from '~types';
-import { getDomainDatabaseId, getVotingClient, verbose } from '~utils';
+import {
+  ColonyMetadata,
+  ColonyOperations,
+  MotionQuery,
+  MotionVote,
+  StakerReward,
+} from '~types';
+import {
+  getColonyFromDB,
+  getDomainDatabaseId,
+  getExistingTokenAddresses,
+  getVotingClient,
+  updateColonyTokens,
+  verbose,
+} from '~utils';
 import networkClient from '~networkClient';
-import { mutate } from '~amplifyClient';
+import { query, mutate } from '~amplifyClient';
 
 export const getStakerReward = async (
   motionId: string,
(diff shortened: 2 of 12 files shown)


willm30:
Didn't mean to request review...

src/handlers/motions/motionCreated/handlers/editColony.ts
```diff
@@ -0,0 +1,19 @@
...
+  { name }: TransactionDescription,
+): Promise<void> => {
+  const { transactionHash, contractAddress: colonyAddress } = event;
+  const pendingColonyMetadataId = getPendingMetadataDatabaseId(
+    colonyAddress,
+    transactionHash,
+  );
+  await createMotionInDB(event, {
+    type: motionNameMapping[name],
+    pendingColonyMetadataId,
```
willm30, line 17: Here we're following the same pattern as with Domains. We attach the "pending" metadata to the action, so it can be retreived when the motion is finalized.

src/handlers/motions/motionFinalized/helpers.ts
```diff
@@ -96,3 +116,84 @@ export const linkPendingDomainMetadataWithDomain = async (action: string, colony
...
+  } = pendingColonyMetadata.changelog?.[0] ?? {};
+
+  const updatedMetadata = {
+    ...currentColonyMetadata,
+  };
+
+  /*
+   * Here, we update metadata as granularly as possible so that we don't overwrite state changes that occured
+   * after this motion was created.
+   */
```
willm30, line 153: As the comment suggests, this should fix the problem of one motion overwriting the state changes of another motion that was created after it, but finalized before it is finalized. To be tested in #433

src/handlers/motions/motionFinalized/helpers.ts
```diff
@@ -96,3 +116,84 @@ export const linkPendingDomainMetadataWithDomain = async (action: string, colony
...
+  }
+
+  if (haveTokensChanged && pendingColonyMetadata.modifiedTokenAddresses) {
+    // If tokens have changed, update colony tokens
+    const colony = await getColonyFromDB(colonyAddress);
+
+    if (colony) {
+      const existingTokenAddresses = getExistingTokenAddresses(colony);
+
+      await updateColonyTokens(
```
willm30, line 182: This logic is taken from the `editColony` saga in the CDapp. It's the same, except we are running it only once a successful motion has been finalized.

src/handlers/motions/motionFinalized/helpers.ts
```diff
@@ -96,3 +116,84 @@ export const linkPendingDomainMetadataWithDomain = async (action: string, colony
...
+        colony,
+        existingTokenAddresses,
+        pendingColonyMetadata.modifiedTokenAddresses,
+      );
+    }
+  }
+
+  await mutate('updateColonyMetadata', {
```
willm30, line 190: Finally, we update the metadata changelog with the changes the motion introduced.

src/utils/colonyClient.ts
```diff
@@ -19,3 +21,20 @@ export const getCachedColonyClient = async (
   colonyClientsCache[colonyAddress] = colonyClient;
   return colonyClient;
 };
+
+export const getColonyFromDB = async (
```
willm30, line 25: We fetch the colony from the db here (only a very limited set of fields currently), so we can get the set of tokens that belong to the colony. This is needed in the event we need to remove tokens.

It does not appear to be the case that this can be queried directly (e.g. "getColonyTokensByColonyId"), since we're using a many-to-many relationship to join them. https://docs.amplify.aws/cli/graphql/data-modeling/#many-to-many-relationship

src/utils/tokens.ts
```diff
@@ -7,3 +11,59 @@ export const getColonyTokenAddress = async (
   const tokenAddress = await colonyClient.getToken();
   return tokenAddress;
 };
+
+export const getExistingTokenAddresses = (colony: ColonyQuery): string[] =>
+  colony.tokens?.items
+    .map((tokenItem) => tokenItem?.tokenAddress)
+    .filter((item): item is string => !!item) ?? [];
+
+export const updateColonyTokens = async (
```
willm30, line 20: This is similar in spirit to its CDapp counterpart (can be found in editColony saga), however, more granular, so that we don't overwrite state changes that happened after the current motion was created.

src/handlers/motions/motionFinalized/helpers.ts
willm30: If this works as I imagine, we can refactor how we store Domain metadata to follow the same pattern.

willm30: https://github.com/JoinColony/block-ingestor/issues/76

Julianahlk (approved):
LGTM 👍

src/utils/tokens.ts
jakubcolony: nice one taking this into consideration 👍

jakubcolony (approved):
Prime work here 💯

PR #75, opened 2023-05-16
Add unlock token status logic to unlockToken handler

ArmandoGraterol:


src/graphql/queries/colony.graphql
willm30, line 1: Is the logic here that when a native token gets unlocked once in one colony, it is now unlocked in every colony that is also using that token as its native token?

willm30: Presumably in this case, you can only unlock the token in the colony that created it?

src/handlers/actions/unlockToken.ts
```diff
@@ -14,6 +25,34 @@ export default async (event: ContractEvent): Promise<void> => {
   const colonyClient = await getCachedColonyClient(colonyAddress);
   const tokenAddress = await colonyClient.getToken();
 
+  const { data } =
+    (await query<
```
willm30, line 29: Here, you probably want to include some mechanism to determine if you've hit the implicit limit (100), and continue fetching until you've got them all. 

> For list queries the default number of objects returned is 100. You can override this behavior by setting the limit argument.

https://docs.amplify.aws/cli/graphql/data-modeling/#model-directive


You can imagine a situation where there are 101 colonies using this token as their native token and currently we'll miss that last 1. 

You should be able to follow the same approach we use for pagination, i.e check if there's a nextToken and if so, repeat the query with that next set of responses, building up an array of responses as you go, until there's no nextToken.

And you can increase the limit also to decrease the number of times this loop will need to run. Given that we're not actually paginating anything, I don't see the harm in setting a larger limit (e.g 1000).

src/graphql/queries/colony.graphql
ArmandoGraterol: That's correct! That is why if you didn't create your colony's native token, you shouldn't see be able to see the "Unlock Token" nor the "Mint Tokens" dialog menu entry in the UAC as the contracts will stop you from minting or unlocking it anyways.

src/handlers/actions/unlockToken.ts
willm30: @ArmandoGraterol Looks good, I would just check the return value of the query function. Internally it accesses the actual data, so I would expect you to be able to write:

```
const {items: colonies, nextToken} = await query()...
```

willm30: And maybe leave a short comment in the code reminding future us what we're up to here.

willm30 (approved):
Looks good from a code perspective (left one comment re the value of the query return). 

I will leave testing to qa as it's a bit tricky to mix and match branches / environments.

jakubcolony (changes requested):
Nice, the UI works as expected but if we need to get a colony/colonies by native token id we should set an index on that field in the CDapp's schema. This would also mitigate the issue with hitting the limit (it'd only become problematic if there are more than 1000 colonies with the same native token, which seems unlikely, whereas now we'll need to start making multiple queries as soon as there are more than 1000 colonies in total).

Something like that should work:

```gql
nativeTokenId: ID! @index(name: "byNativeTokenId", queryField: "getColonyByNativeTokenId")
nativeToken: Token! @hasOne(fields: ["nativeTokenId"])
```

ArmandoGraterol:
@jakubcolony Uploaded https://github.com/JoinColony/colonyCDapp/pull/576, let me know if that's what you had in mind

jakubcolony (approved):
Very nice 💯

willm30:
@ArmandoGraterol Think this is good to merge now?

PR #77, opened 2023-05-18
Refactor updating domain metadata on motion finalized

willm30:
This pr ensures that only the domain metadata that was changed by a motion gets updated in the database. 

Previously, the behaviour was updating the entire domain metadata state with the entire pendingMetadata state, which meant that any changes occuring to the domain metadata after the motion was created, even ones that were not explicitly changed by the motion, would be overwritten.

For example:

Motion 1 created: changes color from to blue to green
Motion 2 created: changes description. Doesn't change color at all.
Motion 1 finalizes, and color changes to green.
Motion 2 finalizes, and changes color back to blue, because that was the color when Motion 2 was created. Expected behaviour would be to leave color alone, because the motion didn't change the color of the domain, only the description.

To test:
Run block ingestor separately on this commit.
Go to CDapp and test the flow above. 

Now, finalizing motion 2 shouldn't affect the domain's color. The final state should be color green, with the updated description (i.e. both changes are persisted simulataneously).

### DIFF
--- a/src/handlers/motions/motionFinalized/helpers.ts
+++ b/src/handlers/motions/motionFinalized/helpers.ts
@@ -4,7 +4,7 @@ import { TransactionDescription } from 'ethers/lib/utils';
 import {
   ColonyMetadata,
   ColonyOperations,
-  MotionQuery,
+  DomainMetadata,
   MotionVote,
   StakerReward,
 } from '~types';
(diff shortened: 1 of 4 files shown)


src/handlers/motions/motionFinalized/helpers.ts
```diff
@@ -101,17 +101,63 @@ export const linkPendingDomainMetadataWithDomain = async (
...
+      oldColor,
+      oldDescription,
+      oldName,
+    } = pendingChangelog[pendingChangelog.length - 1] ?? {};
+
+    const hasColorChanged = newColor !== oldColor;
+    const hasDescriptionChanged = newDescription !== oldDescription;
+    const hasNameChanged = newName !== oldName;
+
+    if (hasColorChanged) {
```
willm30, line 145: only update the metadata if it was updated in the motion that's being finalized

jakubcolony (approved):
Very nice 👍

PR #82, opened 2023-05-29
Add unclaimed stakes support

willm30:
This PR adds block ingestor support for the Stakes Tab in the Token Activation Popover.

Test with: https://github.com/JoinColony/colonyCDapp/pull/591

PR #83, opened 2023-05-30
Refactor event queue to run in sequence

willm30:
Currently, the block ingestor processes events concurrently. I.e. It doesn't wait for Event 1 to finish before processing Event 2. 

This is a problem in the case that two events depend on the same state. Event 2 will overwrite the changes made by Event 1. 

This PR makes the eventQueue wait for each event to finish being processed before it starts processing another. I have tested with https://github.com/JoinColony/block-ingestor/pull/82, and haven't noticed any problems. 

I recommend testing via that pr also. All events should be handled as expected (basically as before, but sequential).

## Example:

Two users claim their stakes on the same motion at the same time.

Both claims are processed with the current state of the motion, which is:

User 1 claimed: false,
User 2 claimed: false

User 1's claim is processed first, and updates the motion state to: 

User 1 claimed: true
User 2 claimed: false

Then User 2's claim is processed, using the original -- not the updated -- state. It updates the state to:

User 1 claimed: false,
User 2 claimed: true

Because Event 2 didn't wait for Event 1 to finish and use the updated state as the basis for its own state changes, we have overwritten the changes made by User 1. If it waits, the final state looks like:

User 1 claimed: true,
User 2 claimed: true

### DIFF
--- a/src/eventQueue.ts
+++ b/src/eventQueue.ts
@@ -9,65 +9,44 @@ dotenv.config();
 
 class EventQueue extends EventEmitter {
   queue: ContractEvent[];
+  processing: boolean;
 
   constructor(options?: Record<string, unknown>) {
     super(options);
     this.queue = [];
+    this.processing = false;
+  }
+
+  async processEvents(): Promise<void> {
+    this.processing = true;
+    while (this.queue.length) {
+      const event = this.queue.shift();
+      if (event) {
+        verbose('Processing event:', event.signature);
+        await eventProcessor(event);
+      }
+    }
+
+    this.processing = false;
   }
 }
 
 const eventQueue = new EventQueue();
 
 /*
- * Process a individual queue event
- */
-eventQueue.on(
-  QueueEvents.ProcessEvent,
-  function (this: EventQueue, event: ContractEvent) {
-    verbose('Processing event:', event.signature);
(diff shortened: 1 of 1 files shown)


src/eventQueue.ts
```diff
@@ -9,65 +9,44 @@ dotenv.config();
...
- *
- * Start processing events from the queue
- */
-eventQueue.on(QueueEvents.ProcessEvents, function (this: EventQueue) {
-  verbose('Processing', this.queue.length, 'events in queue');
-  for (let index = 0; index < this.queue.length; index += 1) {
-    this.emit(QueueEvents.ProcessEvent, this.queue.shift());
+  async processEvents(): Promise<void> {
+    this.processing = true;
+    while (this.queue.length) {
```
willm30, line 22: Because the reference to `this.queue` is dynamic, it means this loop will respond to all calls to `addEvent` for as long as it is running. This means we only have to call `processEvents` once, and it will process all events in the queue for as long as there are events in the queue, even if events are added after `processEvents` was initially called.

jakubcolony (approved):
Nice one 💯

ArmandoGraterol (approved):
Tested with the claim PR and it looks good to me 👍

PR #85, opened 2023-06-07
Filter out 0 colony claims

ArmandoGraterol:
Instructions to test are in https://github.com/JoinColony/colonyCDapp/pull/627


Julianahlk (approved):
LGTM 👍

jakubcolony (approved):
All good here too, maybe leaving a comment as to why we're filtering out 0 could be useful

PR #90, opened 2023-06-14
Add extension/colony permission action event filtering

ArmandoGraterol:


src/utils/actions/writeAction.ts
```diff
@@ -22,18 +27,41 @@ export const writeActionFromEvent = async (
...
+  const shouldShowAction = !(
+    extensionAddresses.find(isAddressInitiatorOrRecipient) ??
+    isAddressInitiatorOrRecipient(networkClient.address)
   );
 
-  verbose('Action', actionType, 'took place in Colony:', colonyAddress);
-  await mutate<CreateColonyActionMutation, CreateColonyActionMutationVariables>(
-    CreateColonyActionDocument,
-    {
+  if (shouldShowAction) {
```
willm30, line 51: Our current approach for filtering the actions list is to set the `showInActionsList` flag, instead of conditionally adding the action to the db. 

It'd be good if we could follow that pattern here, i.e. move the logic inside the `showActionInActionsList` function.

ArmandoGraterol: Are we getting something out of storing an action that we are never going to show? The way I see actions with `showInActionsList` set to `false` is that they _at some point_ could be displayed in the list.

willm30: It is intentional, as far as I'm aware. We're using the db to keep track of everything, regardless of whether it gets shown to the user. For example, we don't remove motions that belong to an uninstalled voting rep extn. 

It might prove useful for things like recovering lost data. I'd check with Raul for full rationale.

ArmandoGraterol:
@jakubcolony 
> * Can this check be done on per-action-type basis instead of for every action? My understanding is that we want to hide (or not store in the DB) permissions-related events

I think that if we go the route of using `showActionInActionsList` as Will suggested above, we could check the `actionFields.type` for permission actions and just filter those.

> * If we're checking for whether the initiator/recipient address matches the extension's address, does it not defeat the point of having `initiatorExtension` and `recipientExtension` fields on the `ColonyAction` model? Seems like they would never get populated anyway.

They are not getting populated manually as you pointed out:
`initiatorExtension: ColonyExtension @hasOne(fields: ["initiatorAddress"])`
`recipientExtension: ColonyExtension @hasOne(fields: ["recipientAddress"])`

They offer more data in the case of the addresses being from extensions, like the `hash`. Like how `recipientColony` is used for example

src/utils/actions/writeAction.ts
ArmandoGraterol: If that's the aim, makes sense to me

jakubcolony:
> I think that if we go the route of using showActionInActionsList as Will suggested above, we could check the actionFields.type for permission actions and just filter those.

Or move the check to the relevant action handler? 

> They are not getting populated manually as you pointed out:
initiatorExtension: ColonyExtension @hasOne(fields: ["initiatorAddress"])
recipientExtension: ColonyExtension @hasOne(fields: ["recipientAddress"])

> They offer more data in the case of the addresses being from extensions, like the hash. Like how recipientColony is used for example

Yes, that's all correct. What I meant though is that if we stop writing any action for which the initiator or recipient address is an extension, these fields will never get populated, rendering them useless. Obviously this doesn't apply if the check only happens for relevant actions, ie. permissions actions

ArmandoGraterol:
@jakubcolony 
> Or move the check to the relevant action handler?

If we use `showInActionsList` to determine if we show/hide actions, wouldn't it be better to group all those checks into `showActionInActionsList`? We can move this check to the handler sure, but then we are going to need to update `writeActionFromEvent` anyways to accommodate for this exception. The exception is us passing `showInActionsList` from the manage permission handler to `writeActionFromEvent` when normally that prop is determined inside it, and I feel like that might complicate the code more than just plugging it in `showActionInActionsList`.

Let me know what you think!

> Yes, that's all correct. What I meant though is that if we stop writing any action for which the initiator or recipient address is an extension, these fields will never get populated, rendering them useless. Obviously this doesn't apply if the check only happens for relevant actions, ie. permissions actions

Ah, I see what you mean now! Yeah I agree, I'm updating the check to only check for the manage permission action type.

src/graphql/queries/extensions.graphql
```diff
@@ -14,3 +14,13 @@ query GetVotingRepInstallations($votingRepHash: String!, $colonyAddress: ID!) {
     }
   }
 }
+
+query GetColonyExtensionsByColonyAddress($colonyAddress: ID!) {
+  listColonyExtensions(
```
willm30, line 19: I think it's better to use `getExtensionByColonyAndHash` here (the hash is optional), since the `list` will list you the first 100 installations across all colonies, and then apply the colony-specific filter, meaning you could end with no results if there are enough colonies and installations in the db. 

Whereas `getExtensionByColonyAndHash` handles the filtering before the results are returned, meaning you're guaranteed to get what you're expecting.

src/utils/actions/writeAction.ts
```diff
@@ -68,8 +73,32 @@ const getVotingReputationInstallations = async (
 
 const showActionInActionsList = async (
   colonyAddress: string,
-  initiatorAddress: string,
+  actionFields: ActionFields,
 ): Promise<boolean> => {
+  const { initiatorAddress, recipientAddress, type } = actionFields;
+
+  if (type === ColonyActionType.SetUserRoles) {
```
willm30, line 80: Could we remove this `if` statement and remove the logic below that only targets actions created by the voting rep extn? 

The logic you've added here will also capture that case, meaning we don't need to run it separately.

willm30 (commented):
Looking good, just a couple of comments/questions

src/utils/actions/writeAction.ts
ArmandoGraterol: Good catch! I didn't realize this

willm30 (approved):
LGTM

PR #91, opened 2023-06-16
Round user stake percentage to 1 if percentage is zero due to rounding

Julianahlk:
For a user stake, there are a few places in the staking UI where a 0% user stake is displayed and this can conflict with the motion stake percentage that always shows a minimum of 1%.

Since its not possible for a user to stake 0% and it visually makes more sense to show 1% since we don't show fractional percentages, this fix makes it so the minimum percent for a user stake is always 1%.

Contributes to colonyCDapp#648


ArmandoGraterol (approved):
Nice work! Looks to be working as expected for me 👍

willm30 (approved):
Nice 👌

willm30:
@Julianahlk I'm noticing that this change has undone https://github.com/JoinColony/block-ingestor/pull/69, [issue](https://github.com/JoinColony/colonyCDapp/issues/509). Would you mind fixing?

PR #94, opened 2023-06-23
Fix: rounding wrong side

Julianahlk:
This fixes an issue that was causes by #91 that undid a fix in #69

This adds back the fix for rounding up to one for wrong stake side.

### DIFF
--- a/src/handlers/motions/motionStaked/helpers.ts
+++ b/src/handlers/motions/motionStaked/helpers.ts
@@ -57,7 +57,7 @@ export const getRemainingStakes = (
 export const getStakePercentage = (
   stake: BigNumber,
   requiredStake: BigNumber,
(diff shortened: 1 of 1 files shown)


src/handlers/motions/motionStaked/helpers.ts
```diff
@@ -55,10 +57,13 @@ export const getRemainingStakes = (
 export const getStakePercentage = (
   stake: BigNumber,
   requiredStake: BigNumber,
+  roundToOne: boolean,
```
willm30, line 60: This could have a default param set to `true` so you don't need to pass it explicitly below

willm30 (approved):
LGTM 👍

ArmandoGraterol (approved):
Good work!

PR #95, opened 2023-06-27
Update env.example with NODE_ENV var

willm30:
Small PR to update env.example since the block ingestor is now looking for the NODE_ENV env var


Julianahlk (approved):
LGTM 👍

ArmandoGraterol (approved):
Nice one!

PR #96, opened 2023-06-28
Add support for displaying staked tokens in Token Activation Popover

willm30:
Note: ~200 lines are from `generated.ts`.

This PR adds support for displaying the amount of tokens a user has staked in a colony to the block ingestor. Test with: https://github.com/JoinColony/colonyCDapp/pull/704

### DIFF
--- a/src/handlers/motions/motionRewardClaimed/helpers.ts
+++ b/src/handlers/motions/motionRewardClaimed/helpers.ts
@@ -1,11 +1,44 @@
-import { mutate } from '~amplifyClient';
+import { BigNumber } from 'ethers';
+import { mutate, query } from '~amplifyClient';
 import {
+  GetColonyStakeDocument,
+  GetColonyStakeQuery,
+  GetColonyStakeQueryVariables,
   StakerReward,
   UpdateColonyDocument,
   UpdateColonyMutation,
   UpdateColonyMutationVariables,
+  UpdateColonyStakeDocument,
+  UpdateColonyStakeMutation,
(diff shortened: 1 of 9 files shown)


src/handlers/motions/motionRewardClaimed/helpers.ts
```diff
@@ -1,11 +1,43 @@
...
+  UpdateUserStakeMutation,
+  UpdateUserStakeMutationVariables,
+  UserStakes,
 } from '~graphql';
-import { getColonyFromDB } from '~utils';
+import { getColonyFromDB, verbose } from '~utils';
+import { getUserStakeId } from '../helpers';
+import { BigNumber } from 'ethers';
+
+export const getUpdatedStakerRewards = (
```
willm30, line 19: This was previously in the main file, moved here

src/handlers/motions/motionRewardClaimed/helpers.ts
```diff
@@ -1,11 +1,44 @@
...
+  UpdateColonyStakeMutation,
+  UpdateColonyStakeMutationVariables,
+  UserStakes,
 } from '~graphql';
-import { getColonyFromDB } from '~utils';
+import { getColonyFromDB, verbose } from '~utils';
+
+import { getColonyStakeId } from '../helpers';
+
+export const getUpdatedStakerRewards = (
```
willm30, line 20: This was previously in motionRewardClaimed.ts, moved here to keep that file neat

PR #104, opened 2023-07-17
Handle expenditure cancellation

jakubcolony:
This PR adds a handler for the `ExpenditureCancelled` event. 

Once you've tested this PR I recommend testing #105 since many testing steps are common.

## Testing
```js

```
2. You can use the following query to verify the expenditure has been stored in the DB (and its status is `DRAFT`):
```gql
query ListExpenditures {
  listExpenditures {
    items {
      id
      status
    }
  }
}
```
3. Cancel the expenditure with the following command:
```js
c.cancelExpenditure(1)
```
4. Using the query above, verify the status has updated to `CANCELLED`.

### DIFF
--- a/src/eventListeners/colony.ts
+++ b/src/eventListeners/colony.ts
@@ -78,6 +78,7 @@ export const setupListenersForColony = (colonyAddress: string): void => {
     ContractEventsSignatures.ExpenditureRecipientSet,
     ContractEventsSignatures.ExpenditurePayoutSet,
     ContractEventsSignatures.ExpenditureLocked,
+    ContractEventsSignatures.ExpenditureCancelled,
   ];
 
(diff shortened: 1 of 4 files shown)


src/eventListeners/colony.ts
```diff
@@ -78,6 +78,7 @@ export const setupListenersForColony = (colonyAddress: string): void => {
     ContractEventsSignatures.ExpenditureRecipientSet,
     ContractEventsSignatures.ExpenditurePayoutSet,
     ContractEventsSignatures.ExpenditureLocked,
+    ContractEventsSignatures.ExpenditureCancelled,
```
rdig, line 81: I assume the actual event `enum` was added in some other PR

rdig (approved):
Look all good :+1:

PR #105, opened 2023-07-17
Handle expenditure finalization

jakubcolony:
This PR adds a handler for the `ExpenditureFinalized` event. 

## Testing
```js

```
2. You can use the following query to verify the expenditure has been stored in the DB (and its status is `DRAFT`):
```gql
query ListExpenditures {
  listExpenditures {
    items {
      id
      status
    }
  }
}
```
3. Finalize the expenditure with the following command:
```js
c.finalizeExpenditure(1)
```
4. Using the query above, verify the status has updated to `FINALIZED`.


rdig (approved):
Same here, since it's basically #104 but with finalized instead of cancelled

PR #106, opened 2023-07-18
Handle expenditure transfer

jakubcolony:
This PR adds a handler for the `ExpenditureTransferred` event. 

## Testing
```js

```
2. You can use the following query to verify the expenditure has been stored in the DB with your wallet address as the `ownerAddress`.
```gql
query ListExpenditures {
  listExpenditures {
    items {
      id
      ownerAddress
    }
  }
}
```
3. Transfer the expenditure ownership with the following command:
```js
c.transferExpenditure(1, '<new owner address>')
```
4. Using the query above, verify the `ownerAddress` has been updated to the new owner address.


rdig (approved):
Slightly different that the other too, but still a short and concise handler. :+1:

PR #107, opened 2023-07-18
Handle setting expenditure claim delay

jakubcolony:
This PR adds handlers for `ExpenditureGlobalClaimDelaySet` and `ExpenditureClaimDelaySet` events.

### Common step
```js

```

### ExpenditureGlobalClaimDelaySet
1. Use the following query to verify there's no global claim delay set:
```gql
query GetColony {
  getColony(id: "<your colony address>") {
    expendituresGlobalClaimDelay
  }
}
```
2. In truffle console, set the global claim delay:
```js
c.setDefaultGlobalClaimDelay(100) // 100 seconds
```
3. Verify the claim delay has been correctly stored on the colony object.

### ExpenditureClaimDelaySet
1. Run the following commands in truffle:
```js
// Set recipient address for expenditure slot 1
c.setExpenditureRecipients(1, 1, '<recipient address>')
// Set the payout amount for slot 1
c.setExpenditurePayout(1, 1, '<token address>', 500)
// Set the claim delay for slot 1 
c.setExpenditureClaimDelay(1, 1, 100) // 100 seconds
```
2. Using the following query, verify the data has been correctly stored in the DB:
```gql
query ListExpenditures {
  listExpenditures {
    items {
      slots { 
        id
        recipientAddress
        claimDelay
        payouts {
          tokenAddress
          amount
        }
      } 
    }
  }
}
```

### DIFF
--- a/src/handlers/expenditures/expenditureClaimDelaySet.ts
+++ b/src/handlers/expenditures/expenditureClaimDelaySet.ts
@@ -1,0 +1,59 @@
+import { mutate } from '~amplifyClient';
+import {
+  ExpenditureSlot,
+  UpdateExpenditureDocument,
+  UpdateExpenditureMutation,
+  UpdateExpenditureMutationVariables,
+} from '~graphql';
+import { ContractEvent } from '~types';
+import { getExpenditureDatabaseId, output, toNumber, verbose } from '~utils';
+
+import { getExpenditure } from './helpers';
+
+export default async (event: ContractEvent): Promise<void> => {
+  const { contractAddress: colonyAddress } = event;
+  const { expenditureId, slot, claimDelay } = event.args;
+  const convertedExpenditureId = toNumber(expenditureId);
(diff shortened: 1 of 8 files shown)


src/handlers/expenditures/expenditureClaimDelaySet.ts
```diff
@@ -0,0 +1,59 @@
...
+  );
+  const updatedSlot: ExpenditureSlot = {
+    ...existingSlot,
+    id: convertedSlot,
+    claimDelay: convertedClaimDelay,
+  };
+  const updatedSlots = [
+    ...expenditure.slots.filter((slot) => slot.id !== convertedSlot),
+    updatedSlot,
+  ];
```
rdig, line 44: Slots are very hard to understand, luckily we're just dealing with their hash rather than decoding/encoding content.

rdig (approved):
Seems all good

PR #108, opened 2023-07-19
Handle setting expenditure payout modifier

jakubcolony:
This PR adds a handler for ExpenditurePayoutModifierSet event.

## Testing
```js

// Set payout modifier of 100 for slot 1
c.setExpenditurePayoutModifiers(1, [1], [100]) // arguments: expenditure id, array of slot numbers, array of modifiers
```

2. Use the following query to verify the modifier has been stored:
```gql
query ListExpenditures {
  listExpenditures {
    items {
      slots { 
        id
        payoutModifier
      } 
    }
  }
}
```


rdig (approved):
Yep, similar to #107  we're just updating the slot with the payout modifier, rather than the claim delay

PR #109, opened 2023-08-01
Add simple decisions support

willm30:
This PR provides decisions support in the block-ingestor.


rdig (approved):
Yep, seems to be functioning as expected.

Very nice turnaround of Simple Decisions overall! :100:

PR #111, opened 2023-08-02
Add annotations support

willm30:
This PR adds annotations support to the block ingestor.

Test with: https://github.com/JoinColony/colonyCDapp/pull/909

### DIFF
--- a/src/handlers/actions/annotateTransaction.ts
+++ b/src/handlers/actions/annotateTransaction.ts
@@ -1,16 +1,80 @@
-import { mutate } from '~amplifyClient';
+import { mutate, query } from '~amplifyClient';
 import {
+  GetActionIdFromAnnotationDocument,
+  GetActionIdFromAnnotationQuery,
+  GetActionIdFromAnnotationQueryVariables,
+  GetMotionIdFromActionDocument,
+  GetMotionIdFromActionQuery,
+  GetMotionIdFromActionQueryVariables,
   UpdateColonyActionDocument,
   UpdateColonyActionMutation,
   UpdateColonyActionMutationVariables,
+  UpdateColonyMotionDocument,
+  UpdateColonyMotionMutation,
+  UpdateColonyMotionMutationVariables,
 } from '~graphql';
 import { ContractEvent } from '~types';
+import { verbose } from '~utils';
 
 export default async ({ args }: ContractEvent): Promise<void> => {
-  const [, actionTxHash] = args;
+  const [, txHash] = args;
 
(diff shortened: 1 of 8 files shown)


src/handlers/actions/annotateTransaction.ts
```diff
@@ -0,0 +1,80 @@
...
+  UpdateColonyActionMutationVariables,
+  UpdateColonyMotionDocument,
+  UpdateColonyMotionMutation,
+  UpdateColonyMotionMutationVariables,
+} from '~graphql';
+import { ContractEvent } from '~types';
+import { verbose } from '~utils';
+
+export default async ({ args }: ContractEvent): Promise<void> => {
+  const [, txHash] = args;
```
rdig, line 20: I think there's two (edge-ish) cases to consider still:
- someone annotates the same transaction before CDapp sends out the "real" annotation, meaning our annotation will be second
- someone annotates the same transaction again

I'm not totally sure on what happens with the first, I think it'll just get assign the `txHash` as the `annotationId`, but then AppSync won't be able to find the reference to it, as the one we've uploaded will have a different id

I think the second case is kinda the same, as far as the end result goes, but with the added issue of being far more easy to replicate and/or trigger. It'll basically overwrite the `annotationId` with an id AppSync won't be able to find.

So, to cover these, I think we need to:
- for the first case, ideally would be to if the user that created the transaction to be annotated matches the user that is annotating the transaction, and only then set it (and this will tie into the next part below)
- for the second, I think we need to check if the action/motion already has a `annotationId` set, before trying to update it again

Makes sense?

rdig (commented):
Good, however I think we still need to cover two extra edge (more or less) case.

I've left a comment below explaining my though process.

src/handlers/actions/annotateTransaction.ts
willm30: @rdig The annotation in the db has the txHash of the transaction it's annotating as its id. This means that in either case, nothing happens. 

If somebody calls `annotateTransaction` for a transaction after it's already been called for the same transaction, this means the txHash on the event will be the same as the first time. So we perform a mutation that's the same as the first time and has no effect.

If somebody calls `annotateTransaction` for a transaction before the CDapp can for the same transaction (unlikely), this will be the same as above: it's just annotating the same transaction again (assuming the corresponding annotation is in the db when they annotate).

If the annotation does not exist in the db, we return early and do nothing. This could happen if somebody is annotating a transaction  via the Dapp or directly on the contract.

thomassdk (approved):
LGTM

PR #116, opened 2023-08-08
Add support for ArbitraryTransaction event

ArmandoGraterol:
This is a pseudo-feature branch used in the CDapp's `port/safe-control` to detect the event emitted after you make a Safe Transaction :tm:

It also provides motion support for the `makeArbitraryTransactions` action, which is the action that we use to send safe transactions.


thomassdk (approved):
LGTM nice work

PR #124, opened 2023-09-03
Cache expenditure payout claimed status

jakubcolony:
https://github.com/JoinColony/colonyCDapp/pull/1036

### DIFF
--- a/src/handlers/expenditures/expenditurePayoutClaimed.ts
+++ b/src/handlers/expenditures/expenditurePayoutClaimed.ts
@@ -1,7 +1,74 @@
 import { ContractEvent } from '~types';
+import { getExpenditureDatabaseId, output, toNumber, verbose } from '~utils';
+
+import { getExpenditureFromDB } from './helpers';
+import {
+  ExpenditurePayout,
+  UpdateExpenditureDocument,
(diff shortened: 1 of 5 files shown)


src/handlers/expenditures/expenditurePayoutClaimed.ts
```diff
@@ -1,7 +1,74 @@
...
+    output(
+      `Could not find expenditure with ID: ${convertedExpenditureId} and colony address: ${colonyAddress} in the db. This is a bug and needs investigating.`,
+    );
+    return;
+  }
+
+  const existingSlotIndex = expenditure.slots.findIndex(
+    (slot) => slot.id === convertedSlotId,
+  );
+  const existingSlot = expenditure.slots[existingSlotIndex];
```
willm30, line 31: Think you could just use `.find` here and below

willm30 (approved):
LGTM

src/handlers/expenditures/expenditurePayoutClaimed.ts
jakubcolony: I need the index though - it's referred to in lines 50 and 58

PR #125, opened 2023-09-04
Handle Staged Expenditure events

jakubcolony:
https://github.com/JoinColony/colonyCDapp/pull/1040


src/eventListeners/extension/index.ts
```diff
@@ -0,0 +1,100 @@
+import { utils } from 'ethers';
```
jakubcolony, line 1: Combined `staked` and `voting` listeners into more generic one for extensions

willm30 (approved):
LGTM

PR #126, opened 2023-09-07
Expenditure Fund Motion support

willm30:
Support for motion funding in block ingestor.

### DIFF
--- a/src/handlers/motions/motionCreated/handlers/moveFunds.ts
+++ b/src/handlers/motions/motionCreated/handlers/moveFunds.ts
@@ -14,20 +14,35 @@ import { createMotionInDB } from '../helpers';
 export const handleMoveFundsMotion = async (
   event: ContractEvent,
   parsedAction: TransactionDescription,
+  gasEstimate: BigNumber,
 ): Promise<void> => {
   const { colonyAddress } = event;
+
   if (!colonyAddress) {
     return;
   }
 
   const { name, args: actionArgs } = parsedAction;
 
-  const [, , , , , fromPot, toPot, amount, tokenAddress] = actionArgs;
-
   const colonyClient = await getCachedColonyClient(colonyAddress);
 
   if (!colonyClient) {
     return;
+  }
+
+  const colonyVersion = await colonyClient.version();
+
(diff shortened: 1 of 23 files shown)


src/handlers/motions/motionCreated/handlers/moveFunds.ts
```diff
@@ -14,22 +14,37 @@ import { createMotionInDB } from '../helpers';
...
 
+  const colonyVersion = await colonyClient.version();
+
+  // There are two moveFundsBetweenPots actions, one pre colony version 7 and one post.
+  let fromPot: BigNumber,
+    toPot: BigNumber,
+    amount: BigNumber,
+    tokenAddress: string;
+
+  const isOldVersion = colonyVersion.lte(6);
```
willm30, line 41: Fixing an oversight here from when originally wiring in the moveFunds motion

src/handlers/motions/motionCreated/handlers/multicall/fragments.ts
```diff
@@ -0,0 +1,25 @@
...
+  event,
+  gasEstimate,
+}: {
+  event: ContractEvent;
+  args: Result;
+  gasEstimate: string;
+}) => Promise<void>;
+// Here we list tuples with the fragments that we are aware of that use multicall, and the corresponding handler
+// we want to fire if the fragment was included in the multicall event.
+export const multicallFragments: Array<[string, MultiCallHandler]> = [
```
willm30, line 16: As suggested in the commnet,  this permits us to handle all the actions we expect to be wrapped in a mutlicall separately.

src/handlers/motions/motionCreated/handlers/networkUpgrade.ts
```diff
@@ -8,11 +9,13 @@ import { createMotionInDB } from '../helpers';
 export const handleNetworkUpgradeMotion = async (
   event: ContractEvent,
   parsedAction: TransactionDescription,
+  gasEstimate: BigNumber,
```
willm30, line 12: I have changed it so we estimate gas for all motion created actions in the block ingestor. This allows us to increase our estimates for actions we know consume more gas, like multicall

willm30:
Merged by mistake, have reverted

PR #133, opened 2023-09-19
Cancel stake motion

willm30:
https://github.com/JoinColony/colonyCDapp/pull/1087

### DIFF
--- a/src/handlers/motions/motionCreated/handlers/expenditures/cancelStakedExpenditures.ts
+++ b/src/handlers/motions/motionCreated/handlers/expenditures/cancelStakedExpenditures.ts
@@ -1,0 +1,35 @@
+import { BigNumber } from 'ethers';
+import { TransactionDescription } from 'ethers/lib/utils';
+import { ContractEvent, motionNameMapping } from '~types';
+import { createMotionInDB } from '../../helpers';
+import {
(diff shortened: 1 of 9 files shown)


src/handlers/motions/motionCreated/handlers/expenditures/cancelStakedExpenditures.ts
```diff
@@ -0,0 +1,35 @@
...
+): Promise<void> => {
+  const { args, colonyAddress } = event;
+  const [, , , , expenditureId] = actionArgs;
+  const [, , domainId] = args;
+
+  if (!colonyAddress) {
+    return;
+  }
+
+  await createMotionInDB(event, {
```
willm30, line 24: If more data is needed for the new ui, it can be added here.

src/handlers/motions/motionCreated/motionCreated.ts
```diff
@@ -140,10 +148,10 @@ export default async (event: ContractEvent): Promise<void> => {
         break;
       }
 
-      // case ColonyOperations.Multicall: {
```
willm30, line 143: Not sure what happened here

jakubcolony (approved):

PR #134, opened 2023-09-21
Add initiator address fallback for recovery action

willm30:
## Description

This is a bug fix as identified here: https://github.com/JoinColony/colonyCDapp/issues/1008

I believe it's caused by setting the recovery role. The recovery role event is handled by the same handler as all other roles, but it has a different event signature: specifically, it doesn't include the `msg.sender`. We were using this as the `initiatorAddress` on the action, but since that field is non-nullable, the mutation fails. 

## Testing

Confirm you're not experiencing the same issues as colonyCDapp#1008 and that the action is created for all permutations of setting permissions.

Resolves  https://github.com/JoinColony/colonyCDapp/issues/1008

### DIFF
--- a/src/handlers/actions/managePermissions.ts
+++ b/src/handlers/actions/managePermissions.ts
@@ -20,11 +20,11 @@ import {
   UpdateColonyRoleDocument,
   ColonyActionType,
 } from '~graphql';
+import provider from '~provider';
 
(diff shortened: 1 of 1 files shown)


rdig (approved):
:+1: 

Just noting that will replace the PR and the fix brought in by JoinColony/colonyCDapp#1097

PR #135, opened 2023-09-21
Fix arbitrary reputation update handler

willm30:
This PR fixes a bug that causes the block ingestor to crash. 

In the arbitrary reputation update handler, we were using event logs to find a domain instead of the db. Lots of logs lead to an rpc timeout, leading to the block ingestor crashing.

### DIFF
--- a/src/handlers/actions/emitDomainReputation.ts
+++ b/src/handlers/actions/emitDomainReputation.ts
@@ -49,7 +49,7 @@ export default async (event: ContractEvent): Promise<void> => {
 
   if (!domain) {
     verbose(
-      'Not acting upon the ArbitraryReputationUpdate event as a domain with the skillId was not found',
+      `Not acting upon the ArbitraryReputationUpdate event as a domain with the skillId ${skillId} was not found`,
     );
     return;
   }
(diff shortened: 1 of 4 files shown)


src/handlers/actions/emitDomainReputation.ts
```diff
@@ -41,23 +25,35 @@ export default async (event: ContractEvent): Promise<void> => {
...
+  let domain;
+  let nextToken: string | undefined;
+  /*
+   * Search for domain with skillId from Colony's domains. We paginate the query
+   * just in case there are more domains than the query limit. We can stop searching
+   * if we find a domain.
+   */
+  do {
+    const { data } =
+      (await query<GetColonyQuery, GetColonyQueryVariables>(GetColonyDocument, {
```
willm30, line 37: All I'm doing here is querying the colony, and filtering its domains by the skillId in the event. 

The reason it's inside a loop is in case there are more domains than the query limit (I've set it manually to 1,000), we won't get all the domains unless we paginate the query. Pretty defensive, but there just in case.

rdig (approved):
This is very nice. A better implementation of what I did with chain events in the Dapp. :100: 

_PS: Sorry for the late review._

PR #137, opened 2023-09-27
Add handler for release expenditure stage motion

ArmandoGraterol:
CDapp counterpart: https://github.com/JoinColony/colonyCDapp/pull/1088

Contributes to https://github.com/JoinColony/colonyCDapp/issues/1062


willm30:
>  is it normal for the gas estimate of motionCreated to fail most of the time

Does it fail for all motions or just the motion you're testing? I also noticed a tendency for it to fail when implementing the other expenditure motion, but not generally (i.e. for other actions).

ArmandoGraterol:
> > is it normal for the gas estimate of motionCreated to fail most of the time
> 
> Does it fail for all motions or just the motion you're testing? I also noticed a tendency for it to fail when implementing the other expenditure motion, but not generally (i.e. for other actions).

It seems like it's just the motion to release the payment so I'll take a closer look at the block ingestor to see what's up.

jakubcolony (approved):
LGTM 👍

PR #144, opened 2023-11-13
Show Payment actions incoming funds

jakubcolony:
This resolves https://github.com/JoinColony/colonyCDapp/issues/991 which is needed to display the Payment actions fund claims after contracts update.

For testing, create a payment action with a colony as a recipient and confirm the incoming funds are showing up.

### DIFF
--- a/src/utils/isColonyAddress.ts
+++ b/src/utils/isColonyAddress.ts
@@ -1,0 +1,9 @@
+import networkClient from '~networkClient';
+
+/**
+ * Function accepting a contract address and returning true if it belongs to a colony
+ * It calls the `isColony` function on the network client
+ */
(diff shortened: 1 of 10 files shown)


src/utils/isColonyAddress.ts
thomassdk, line 1: Out of curiosity is this faster than using the graphql query `getColonyByAddress`?

thomassdk (approved):
LGTM :+1:

src/utils/isColonyAddress.ts
jakubcolony: I haven't measured it but an outright downside to the query is that it won't work with colonies created on-chain directly.

PR #145, opened 2023-11-19
Ensure block timestamp is saved as motion createdAt date

jakubcolony:
This PR fixes an issue where the motion's `createdAt` date is set to the system current time rather than the block timestamp.


rdig (approved):
Not really a fan of using ISOString, but since we're using it all over the block ingestor at this point, might as well keep it consistent

PR #147, opened 2023-12-01
Throw error when failed to get a block

jakubcolony:
If error is not thrown, block ingestor hangs and needs to be manually restarted


ArmandoGraterol (approved):
LGTM 👍

area (approved):

PR #151, opened 2023-12-07
Create colonies in the DB from the Ingestor side

rdig:
This is the other side fo the JoinColony/colonyCDapp#1513 refactor, the part that listens for the `ColonyAdded` event, extracts data from it, then fires off the `createUniqueColony` mutation (via lamba)

The basic logic is that the client will send out the create colony transaction, as soon as it's generated (before even being mined) we get the transaction's hash and use that to send a mutation to the database containing the upcoming colony's metadata.

Here, we pick up the colony created event, extract required data from it, then trigger a mutation (via a lambda). This lambda is quite complex as it encapsulates everything needed to create a colony, it's metadata, it's token, and it's domains.

_Note: have a look at JoinColony/colonyCDapp#1513 for a schema that shows a overview of this logic_

Contributes to JoinColony/colonyCDapp#1574


rdig:
This is back to draft until we figure out the Token Authority set race condition

src/eventListeners/colony.ts
```diff
@@ -96,5 +108,23 @@ export const setupListenersForColony = (colonyAddress: string): void => {
...
+   *
+   * Currently it works as we only care about one specific event which only happens
+   * on colony creation (Transfer doesn't count as it's being handled differently)
+   */
+  tokenEvents.forEach((eventSignature) => {
+    // This needs to be handled individually because transfer events are frequent and heavy
+    if (eventSignature === ContractEventsSignatures.Transfer) {
+      return addTokenTransferEventListener(eventSignature, colonyAddress);
+    }
+    return addTokenEventListener(eventSignature, tokenAddress);
```
area, line 128: As discussed on a call, currently this potentially has quite poor performance if multiple colonies are using the same token and a `LogSetAuthority` event is fired - there will be multiple handlers - one created for each colony - and each call with update the appropriate values of all relevant colonies.

As further discussed on the call, using our [existing list of handlers](https://github.com/JoinColony/block-ingestor/blob/master/src/eventListeners/eventListeners.ts#L5) might be a quick way to improve this, checking if we have a relevant handler already.

src/handlers/colonies/colonyAdded.ts
```diff
@@ -82,8 +89,39 @@ export default async (event: ContractEvent): Promise<void> => {
...
+    >(CreateUniqueColonyDocument, {
+      input: {
+        colonyAddress: utils.getAddress(colonyAddress),
+        tokenAddress: utils.getAddress(tokenAddress),
+        transactionHash,
+        initiatorAddress: utils.getAddress(colonyFounderAddress),
+      },
+    });
+  }
+
```
jakubcolony, line 122: I thought you mentioned deleting the temporary metadata from the DB, does that happen elsewhere or not at all yet?

src/handlers/expenditures/helpers/getExpenditure.ts
```diff
@@ -34,15 +29,11 @@ export const getExpenditureFromDB = async (
...
-  AnyColonyClient,
-  ColonyClientV1 | ColonyClientV2 | ColonyClientV3
->;
+type GetExpenditureSupportedColonyClient = AnyColonyClient;
 const isGetExpenditureSupported = (
   colonyClient: AnyColonyClient,
 ): colonyClient is GetExpenditureSupportedColonyClient =>
-  (colonyClient as GetExpenditureSupportedColonyClient).getExpenditure !==
-  undefined;
+  colonyClient.getExpenditure !== undefined;
```
jakubcolony, line 36: You can get rid of the type and the `colonyClient is GetExpenditureSupportedColonyClient`, it doesn't really do anything following colony-js update

src/handlers/tokens/setTokenAuthority.ts
```diff
@@ -0,0 +1,64 @@
...
+export default async (event: ContractEvent): Promise<void> => {
+  const { contractAddress, signature } = event;
+
+  const client = await getCachedTokenClient(contractAddress);
+
+  if (!client) {
+    throw new Error(
+      `Cannot get token client for ${contractAddress} when handling ${signature}`,
+    );
+  }
```
jakubcolony, line 19: This might cause block-ingestor restart loop until a client is found - after crashing and restarting it will attempt to handle this event again I think

It may be the expected behaviour tho, so just checking

src/handlers/tokens/setTokenAuthority.ts
```diff
@@ -0,0 +1,64 @@
...
+      // note that we can't be sure if the token even has the mint function,
+      // or if it does, what signature does it have
+      try {
+        // @ts-expect-error
+        await client.estimateGas.mint(1, { from: colonyAddress });
+        mintable = true;
+      } catch (error) {
+        // silent
+      }
+    }
```
jakubcolony, line 57: Will it actually run async if you don't use something like `Promise.all`?

src/utils/domains.ts
```diff
@@ -21,11 +15,8 @@ export const getPendingMetadataDatabaseId = (
...
-type SupportedColonyClient = Exclude<
-  AnyColonyClient,
-  ColonyClientV1 | ColonyClientV2 | ColonyClientV3 | ColonyClientV4
->;
+type SupportedColonyClient = AnyColonyClient;
 export const isDomainFromFundingPotSupported = (
   colonyClient: AnyColonyClient,
 ): colonyClient is SupportedColonyClient =>
-  (colonyClient as SupportedColonyClient).getDomainFromFundingPot !== undefined;
+  colonyClient.getDomainFromFundingPot !== undefined;
```
jakubcolony, line 22: Same as above

src/utils/tokens.ts
```diff
@@ -93,3 +101,77 @@ export const updateColonyTokens = async (
     }
   });
 };
+
+// Fetch all colonies that share the same native token
+export const fetchColoniesByNativeToken = async (
+  tokenAddress: string,
+): Promise<Array<Pick<FullColony, 'id' | 'status'>>> => {
```
jakubcolony, line 108: Alternatively this could be typed as:
```
Array<
    NonNullable<
      NonNullable<
        GetColonyByNativeTokenIdQuery['getColoniesByNativeTokenId']
      >['items'][number]
    >
  >
```
which is ugly but it would tie it to the actual query, right now this won't be updated if the queried fields change. Or you could create a new fragment for it

jakubcolony (changes requested):
I left a few comments but overall this looks & works great, nice work

src/eventListeners/colony.ts
```diff
@@ -96,5 +108,23 @@ export const setupListenersForColony = (colonyAddress: string): void => {
...
+  });
```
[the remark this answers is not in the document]
rdig: I agree with you that for now, this is overkill, however my point for having this in is to not have to deal with this in the future, when more token based events will be required. Just add them in, and this will do the rest.

I see both side of the argument as having validity here.

src/eventListeners/token.ts
```diff
@@ -16,3 +16,15 @@ export const addTokenEventListener = (
...
+  eventSignature: ContractEventsSignatures,
+  address: string,
+): void => {
+  addEventListener({
+    type: EventListenerType.Token,
+    address,
+    eventSignature,
+    topics: [utils.id(eventSignature)],
+  });
+};
```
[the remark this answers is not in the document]
rdig: I thought that would be too "dirty", and choose to add a new one, however I'll give it a try again and see if I can't come up with something better.

src/handlers/colonies/colonyAdded.ts
rdig: That's being done in the actual `createUniqueColony` lambda: https://github.com/JoinColony/colonyCDapp/blob/refactor/colony-creation/amplify/backend/function/createUniqueColony/src/index.js#L298-L306

I agree with you now, that all the content of that lambda should be eventually moved to the block ingestor _(and the lambda removed)_. But that's a refactor for another time.

src/handlers/tokens/setTokenAuthority.ts
rdig: Makes sense actually. Removed it and just put in a check to handle the event if the client is found. That will still have the desired effect, while preventing the type of behavior you are describing.

src/handlers/tokens/setTokenAuthority.ts
rdig: You are correct that this might be a problem normally _(even with `Promise.all`)_ as the way `forEach` is designed, it can't handle promises properly _(especially rejections)_

It kinda works for our case, as we don't really care about the result, since each colony is independent and can be updated whenever. 

The real problem here begins if there's ever an error, as `forEach` won't catch it properly and it has the potential of crashing the node process.

As such, I've refactored both this, and the other instance that's using this partern to `for...of`

src/eventListeners/token.ts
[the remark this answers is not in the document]
rdig: This was refactored to both include your requirements as well as Alex's

jakubcolony (approved):
All good here, just a little suggestion about the token listeners

src/eventListeners/token.ts
```diff
@@ -1,18 +1,41 @@
...
+      listener.type === EventListenerType.Token &&
+      listener.address === recipientAddress,
+  );
+
+  const tokenListener: EventListener = {
     type: EventListenerType.Token,
     eventSignature,
-    topics: [utils.id(eventSignature), null, recipientAddress],
-  });
+    address: recipientAddress,
```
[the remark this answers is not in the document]
rdig: > The address property here should be set to tokenAddress, and the if below could depend on whether recipientAddress is defined, as opposed to event signature

That's not correct approach, since then you can break this by just passing whatever in in the `recipientAddress`. The "filter by recipient" patter works only in some cases, and we need to correctly control those. 

But overall I think this has become too complicated and I have half a mind to revert my changes back to my initial implementation: one handler for Transfer events, one handler for every other type of token event

src/eventListeners/token.ts
```diff
@@ -1,18 +1,41 @@
...
+    /*
+     * Filter transfer events for the recipient address
+     */
+    tokenListener.topics.push(null, recipientAddress);
+    delete tokenListener.address;
+  }
+
+  if (!listenerExists) {
+    return addEventListener(tokenListener);
+  }
```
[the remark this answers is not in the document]
rdig: I don't quite understand what you're suggesting here

PR #154, opened 2023-12-21
Read contract data at the time the event happened

rdig:
This PR refactors **all** the contract read calls we make in the block ingestor _(ie: every time we call a method on an instantiated `colonyJS` client)_ by making sure the data we read is historically correct to the time the event actually happened.

This is accomplished by getting the block number of the specific event we handle, and passing it in the [CallOverrides](https://docs.ethers.org/v5/api/contract/contract/#Contract-functionsCall) options object

This way we ensure that every time we read a specific piece of data from the contracts, be it now, or in the past, it will have always the exact same, correct value.

This change in this PR helps avoid various race conditions that will start happening once the block ingestor is under heavier load, as well as giving the block ingestor the ability to be properly sync'd from a past block in time, up to the latest one.

### DIFF
--- a/src/handlers/motions/motionCreated/motionCreated.ts
+++ b/src/handlers/motions/motionCreated/motionCreated.ts
@@ -1,4 +1,5 @@
 import { BigNumber, constants } from 'ethers';
+import { JsonRpcProvider } from '@ethersproject/providers';
 
 import { ColonyOperations, ContractEvent } from '~types';
 import {
@@ -30,6 +31,7 @@ export default async (event: ContractEvent): Promise<void> => {
   const {
     args: { motionId },
     colonyAddress,
+    blockNumber,
   } = event;
 
   if (!colonyAddress) {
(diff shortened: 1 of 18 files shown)


src/handlers/motions/motionCreated/motionCreated.ts
```diff
@@ -58,25 +62,30 @@ export default async (event: ContractEvent): Promise<void> => {
...
-        /*
-         * If the motion target is 0x000... then we pass in the colony's address
-         */
-        motion.altTarget === constants.AddressZero
-          ? colonyClient.address
-          : motion.altTarget,
-      data: motion.action,
-    });
+  const estimateMotionGas = async (): Promise<string> =>
+    await (colonyClient.provider as JsonRpcProvider).send('eth_estimateGas', [
```
jakubcolony, line 66: Is this worth a comment as to why the casting is necessary?

jakubcolony (approved):
All looks good, nice work 👍 I'll keep an eye out on that in the future

PR #155, opened 2023-12-21
Increase max retries for colony metadata query

jakubcolony:
I noticed the create data script failing with the max retries set to 3, this gives it a better chance to catch up.


rdig (approved):
Looks all good

PR #166, opened 2024-01-31
[Verified members] feat: block ingestor can now catch MetadataDelta events and parse operations, implement add verified members

bassgeta:
## Description
Added a new handler for the `ColonyMetadataDelta` event. 
The handler tries to parse the event via type guards, it currently only has support for adding verified members.
Using this approach it should be easy to implement removing of verified members.

## How to test
Easiest way is to:
2. spin up the CDapp on [this branch](https://github.com/JoinColony/colonyCDapp/pull/1773).
3. Choose a member that hasn't been verified yet and copy their address 
4. go to the colony dashboard home, where the temporary component for verifying members lives, paste the address and click `Verify`
image
5. Check logs to see that block-ingestor processed the transaction
6. Use GraphQL query `getVerifiedMembersByColony` to check which users are added. The user you have pasted in above should show up in the `items` array

Resolves https://github.com/JoinColony/block-ingestor/issues/160

### DIFF
--- a/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
+++ b/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
@@ -1,3 +1,4 @@
+import { utils } from 'ethers';
 import { mutate, query } from '~amplifyClient';
 import {
   ColonyActionType,
@@ -9,37 +10,26 @@ import {
   GetVerifiedMemberQueryVariables,
 } from '~graphql';
 import { ContractEvent } from '~types';
-import { verbose, writeActionFromEvent } from '~utils';
-import {
-  AddVerifiedMembersOperation,
-  MetadataDeltaOperationType,
-  MetadataDeltaOperation,
-} from '../types';
-
-const isAddVerifiedMembersOperation = (
-  operation: MetadataDeltaOperation,
-): operation is AddVerifiedMembersOperation => {
-  return (
-    operation.type === MetadataDeltaOperationType.ADD_VERIFIED_MEMBERS &&
-    operation.payload !== undefined &&
-    Array.isArray(operation.payload)
-  );
-};
+import { writeActionFromEvent } from '~utils';
+import { AddVerifiedMembersOperation } from '../types';
 
 export const handleAddVerifiedMembers = async (
   event: ContractEvent,
-  operation: MetadataDeltaOperation,
+  operation: AddVerifiedMembersOperation,
 ): Promise<void> => {
-  if (!isAddVerifiedMembersOperation(operation)) {
-    verbose('Malformed AddVerifiedMembers operation: ', operation);
-    return;
-  }
-
   const { contractAddress: colonyAddress } = event;
   const { agent: initiatorAddress } = event.args;
 
-  await Promise.all(
-    operation.payload.map(async (userAddress) => {
+  await Promise.allSettled(
+    operation.payload.map(async (address) => {
+      let userAddress;
+      try {
+        userAddress = utils.getAddress(address);
+      } catch (error) {
+        // Bail if it's not an address
+        return;
+      }
+
       const item = await query<
         GetVerifiedMemberQuery,
         GetVerifiedMemberQueryVariables
@@ -47,10 +37,9 @@ export const handleAddVerifiedMembers = async (
 
       const verifiedMemberData = item?.data?.getVerifiedMember;
 
+      // If user is already verified, don't verify them again
       if (verifiedMemberData !== undefined && verifiedMemberData !== null) {
-        throw new Error(
-          `User ${userAddress} already a verified member of the colony ${colonyAddress}`,
-        );
+        return;
       }
 
       await mutate<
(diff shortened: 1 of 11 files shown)


src/handlers/metadataDelta/metadataDelta.ts
```diff
@@ -0,0 +1,71 @@
...
+  isAddVerifiedMembersOperation,
+  isMetadataDeltaOperation,
+  parseOperation,
+} from './utils';
+import { mutate, query } from '~amplifyClient';
+
+export default async (event: ContractEvent): Promise<void> => {
+  const { contractAddress: colonyAddress } = event;
+  const { agent: initiatorAddress } = event.args;
+  const operationString = event.args[1];
```
jakubcolony, line 22: You could access `operationString` by `event.args.metadata` if I'm not mistaken

src/handlers/metadataDelta/metadataDelta.ts
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  const operation = parseOperation(operationString);
+
+  if (!isMetadataDeltaOperation(operation)) {
+    verbose(
+      'Operation does not conform to MetadataDeltaOperation type: ',
+      operation,
+    );
+    throw new Error('Unknown operation format');
```
jakubcolony, line 35: As discussed, we shouldn't throw errors inside handlers (it will cause a restart loop)

src/handlers/metadataDelta/utils.ts
```diff
@@ -0,0 +1,42 @@
...
+): operation is MetadataDeltaOperation => {
+  return (
+    typeof operation === 'object' &&
+    operation !== null &&
+    operation.type !== undefined &&
+    operation.entity !== undefined &&
+    operation.colonyAddress !== undefined &&
+    typeof operation.colonyAddress === 'string' &&
+    operation.members !== undefined &&
+    Array.isArray(operation.members)
```
jakubcolony, line 29: It looks like there's a conflict between trying to make this generic (operating on an arbitrary Metadata Delta Operation) and being specific to verified members (sounds possible that other operations won't have members field?)

Would it not be easier if your general Metadata Delta Operation would only be concerned with type (Add Members, Remove Members) and payload specific for a type? I also think you don't need colony address here, it's tied to the event itself.

The job of this handler could then be reduced to:
a) check for type of operation, ignore it if it's unknown.
b) forward the operation to an operation-specific handler/util, which could do validation on a per-operation basis.

All operation types could be nicely extracted, similar to how we handle different types of actions in `motionCreated`.

A sample metadata could look like so:
```json
{
  "type": "ADD_VERIFIED_MEMBERS",
  "payload": ["0x...", "0x..."]
}
```

jakubcolony (changes requested):
Awesome! This definitely works and it's great for a first attempt at block-ingestor.

I left you a couple of comments, most of which we already discussed.

Well done again!

src/handlers/metadataDelta/metadataDelta.ts
```diff
@@ -0,0 +1,71 @@
...
+import { mutate, query } from '~amplifyClient';
+
+export default async (event: ContractEvent): Promise<void> => {
+  const { contractAddress: colonyAddress } = event;
+  const { agent: initiatorAddress } = event.args;
+  const operationString = event.args[1];
+
+  if (!operationString) {
+    verbose('Unable to get operation for ColonyMetadataDelta event');
+  }
```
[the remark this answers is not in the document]
jakubcolony: @bassgeta still outstanding, this will get caught further on but might as well stop the execution here

src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
```diff
@@ -0,0 +1,67 @@
...
+    operation.payload !== undefined &&
+    Array.isArray(operation.payload)
+  );
+};
+
+export const handleAddVerifiedMembers = async (
+  event: ContractEvent,
+  operation: MetadataDeltaOperation,
+): Promise<void> => {
+  if (isAddVerifiedMembersOperation(operation)) {
```
jakubcolony, line 33: Please consider adding a log if the payload is badly formatted

jakubcolony (changes requested):
Hey @bassgeta, thanks for making the changes we discussed, I've just done another pass and this looks very good, just left you a couple more comments.

rdig (approved):
Looks good, especially after you addressed Jakub's concerns.

Good to go.

jakubcolony (approved):
Looks great, well done 👍

src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
```diff
@@ -0,0 +1,70 @@
...
+  MetadataDeltaOperation,
+} from '../types';
+
+const isAddVerifiedMembersOperation = (
+  operation: MetadataDeltaOperation,
+): operation is AddVerifiedMembersOperation => {
+  return (
+    operation.type === MetadataDeltaOperationType.ADD_VERIFIED_MEMBERS &&
+    operation.payload !== undefined &&
+    Array.isArray(operation.payload)
```
jakubcolony, line 25: You do the type check in `parseOperation`, so the individual handlers only need to validate the payload, not the entire operation hopefully?

jakubcolony (approved):
Very nice 👌

PR #168, opened 2024-02-01
[Verified members] feat: add support for handling metadata delta motions

### LINKED ISSUE #162: [Verified members] Feature: Block ingestor can parse and handle the Add verified members motion
### Story

The block ingestor must be able to parse the new event which corresponds to adding verified members via motions.
When it's finalized it needs to apply the result (or not apply it if it failed).

### Description

The client will send out a `Metadata delta` motion event where they will pass through the pseudocode language a la `ADD {userAddress} VERIFIED {colonyAddress}` and the block ingestor needs to parse it and handle it.

## Implementation

Create a new handler in the `motions/motionCreated` folder and handle it in `motionCreated.ts` file.
We need to parse the passed in string, so we know which user is being added to which colony.

- [ ] Block ingestor can successfully handle the motion event and discern it being an `Add` event not anything else 
- [ ] Block ingestor successfully creates a motion
- [ ] If the motion passes, we add the verified members via the gql mutation
- [ ] If the motion fails the members don't get added

bassgeta:
## Description
Added a new handler for the motion created - `ColonyMetadataDelta` event. 
The handler tries to parse the event via type guards, it currently only has support for adding verified members.
Using this approach it should be easy to implement removing of verified members.

## How to test
Easiest way is to:
2. spin up the CDapp on [this branch](https://github.com/JoinColony/colonyCDapp/pull/1835).
3. Choose a member that hasn't been verified yet and copy their address 
4. go to the colony dashboard home, where the temporary component for verifying members lives, paste the address and click `Verify via motion`
5. FUlly support the motion
6. run `curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[300],"id":1}' localhost:8545` to forward the time by 5 minutes
7. Finalize the motion, claim rewards, ...
8. Use GraphQL query `getVerifiedMembersByColony` to check which users are added. The user you have pasted in above should show up in the `items` array

You can also check out if it works by rejecting a motion

Resolves https://github.com/JoinColony/block-ingestor/issues/162


src/handlers/motions/motionCreated/handlers/metadataDelta.ts
```diff
@@ -0,0 +1,53 @@
...
+  isAddVerifiedMembersOperation,
+  parseOperation,
+  verbose,
+} from '~utils';
+import { createMotionInDB } from '../helpers';
+
+export const handleMetadataDeltaMotion = async (
+  event: ContractEvent,
+  desc: TransactionDescription,
+  gasEstimate: BigNumber,
```
rdig, line 16: Not related to this PR, but the gas estimate shouldn't need passing around like this.

I guess we'll need to refactor this out at some point.

rdig (approved):
All good

src/utils/metadataDelta/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './utils';
+export * from './types';
```
jakubcolony, line 2: This opens a whole new discussion on block-ingestor folder structure but there's `src/types` folder already and I'm not sure about exporting types from `~utils`.

jakubcolony (approved):
LGTM 👍

jakubcolony (changes requested):

src/handlers/motions/motionCreated/handlers/metadataDelta.ts
```diff
@@ -0,0 +1,53 @@
...
+    const operation = parseOperation(operationString);
+
+    if (operation === null) {
+      return;
+    }
+
+    const pendingColonyMetadataId = getPendingMetadataDatabaseId(
+      colonyAddress,
+      transactionHash,
+    );
```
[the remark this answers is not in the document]
bassgeta: yeah you're right it doesn't make sense, removing it :+1:

jakubcolony (approved):

PR #171, opened 2024-02-12
[Verified members] Implement remove verified members motion

### LINKED ISSUE #163: [Verified members] Feature: Block ingestor can parse and handle the Remove verified members motion
### Story

The block ingestor must be able to parse the new event which corresponds to removing verified members via motions.
When it's finalized it needs to apply the result (or not apply it if it failed).

### Description

The client will send out a `Metadata delta` motion event where they will pass through the pseudocode language a la `REMOVE {userAddress} VERIFIED {colonyAddress}` and the block ingestor needs to parse it and handle it.

## Implementation

Create a new handler in the `motions/motionCreated` folder and handle it in `motionCreated.ts` file.
We need to parse the passed in string, so we know which user is being removed from which colony's verified members.

- [ ] Block ingestor can successfully handle the motion event and discern it being `Remove` event not anything else 
- [ ] Block ingestor successfully creates a motion
- [ ] If the motion passes, we remove the verified members via the gql mutation
- [ ] If the motion fails the members don't get removed

iamsamgibbs:
## Description
Added support for removing verified members to the `ColonyMetadataDelta` event handler.

## How to test
Easiest way is to:
2. spin up the CDapp on [this branch](https://github.com/JoinColony/colonyCDapp/pull/1891).
3. Enable the permissions extension
4. Choose a verified member and copy their address 
5. go to the colony dashboard home, where the temporary component for unverifying members lives, paste the address and click `Unverify via motion`
6. Fully support the motion
7. run `curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[300],"id":1}' localhost:8545` to forward the time by 5 minutes
8. Finalize the motion, claim rewards, ...
9. Use GraphQL query `getVerifiedMembersByColony` to check which users are verified. The user you have pasted in above should no longer appear in the `items` array

You can also check out if it works by rejecting a motion

Resolves #163

### DIFF
--- a/src/handlers/motions/motionCreated/handlers/metadataDelta.ts
+++ b/src/handlers/motions/motionCreated/handlers/metadataDelta.ts
@@ -2,7 +2,12 @@ import { BigNumber } from 'ethers';
 import { TransactionDescription } from 'ethers/lib/utils';
 import { ColonyActionType } from '~graphql';
 import { ContractEvent } from '~types';
-import { isAddVerifiedMembersOperation, parseOperation, verbose } from '~utils';
+import {
+  isAddVerifiedMembersOperation,
+  isRemoveVerifiedMembersOperation,
+  parseOperation,
+  verbose,
+} from '~utils';
 import { createMotionInDB } from '../helpers';
 
 export const handleMetadataDeltaMotion = async (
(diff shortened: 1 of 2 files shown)


rdig (approved):
All good here. Tested and seems to work fine. Nice work.

src/handlers/motions/motionCreated/handlers/metadataDelta.ts
```diff
@@ -47,6 +48,15 @@ export const handleMetadataDeltaMotion = async (
         gasEstimate: gasEstimate.toString(),
       });
     }
+
+    if (isRemoveVerifiedMembersOperation(operation)) {
+      await createMotionInDB(event, {
+        type: ColonyActionType.RemoveVerifiedMembersMotion,
+        members: operation.payload,
+        pendingColonyMetadataId,
```
jakubcolony, line 56: Do you know if this is needed @iamsamgibbs? Not sure if I'm missing something but none of the CDapp sagas seem to be creating the metadata it would link to.

jakubcolony (approved):
Nice work @iamsamgibbs 👍

src/handlers/motions/motionCreated/handlers/metadataDelta.ts
iamsamgibbs: I don't believe it is needed actually, no references to it in the CDapp sagas so I've just removed it from both add and remove and still working fine.

PR #173, opened 2024-02-13
Feature/verified members contributors flag

### LINKED ISSUE #172: [Verified Members] Update the isVerified flag for contributors when verifying / unverifying members
### Story

In the UI, we display verified checkmarks for users which are added as verified. 
We need to realign this according to the verified members rework.

### Description

This should be done by changing the `isVerified` flag on the `ColonyContributor` model.

- [ ] The `isVerified` flag and checkmarks reflect the state of verified members after realign

iamsamgibbs:
Change the isVerified flag on the ColonyContributor model when a member is verified / unverified.

Remove the VerifiedMember model and related queries, using ColonyContributor instead.

---

Please review with https://github.com/JoinColony/colonyCDapp/pull/1898 and find testing notes there.

Resolves #172


bassgeta (approved):
Nice one, works like a charm for me and the API is left unchanged :candy:

PR #175, opened 2024-02-15
Colony balance should include funds currently in expenditures' pots

jakubcolony:
https://github.com/JoinColony/colonyCDapp/pull/1918


bassgeta (approved):
Tried it out on the CDapp and works as you've described :+1:

iamsamgibbs (approved):
Tested with colonyCDapp#1918 and all works as expected, great job!

PR #176, opened 2024-02-16
Advanced Payments API

jakubcolony:
https://github.com/JoinColony/colonyCDapp/pull/1922


src/handlers/motions/motionFinalized/helpers.ts
```diff
@@ -375,3 +390,111 @@ export const updateColonyUnclaimedStakes = async (
...
+  if (!colonyAddress) {
+    output('Colony address missing for ExpenditureStateChanged event');
+    return;
+  }
+
+  // @NOTE: If:
+  // - The value is not 0
+  // - The storage slot is not 26
+  // Then we can assume this state change ain't a stage release
+  if (convertedStorageSlot !== 26 || convertedValue !== 0) {
```
rdig, line 443: I'd maybe add these as `enums`? But maybe that's just me, as I get alarm bells when I see "magic" values  randomly through the code

rdig (approved):
Other than some code styling / structure that I would have maybe written differently, I can't really point to anything that I'd want to see any different in this PR.

All in all, great work here. :100:

src/handlers/motions/motionFinalized/helpers.ts
jakubcolony: Actually this will be replaced by https://github.com/JoinColony/colonyCDapp/issues/2014, but it's a fair point so I added a const with a comment for now.

PR #180, opened 2024-02-26
Feat: Motion handler for release expenditure stage

iamsamgibbs:
See https://github.com/JoinColony/colonyCDapp/pull/1984

### DIFF
--- a/src/handlers/motions/motionFinalized/helpers.ts
+++ b/src/handlers/motions/motionFinalized/helpers.ts
@@ -295,12 +295,12 @@ export const linkPendingMetadata = async (
     colonyAddress,
   );
 
-  const parsedAction = parseAction(action, [
+  const parsedAction = parseAction(action, {
     colonyClient,
     oneTxPaymentClient,
     stakedExpenditureClient,
     stagedExpenditureClient,
-  ]);
+  });
 
   if (!parsedAction) {
     return;
@@ -397,7 +397,7 @@ export const claimExpenditurePayouts = async (
 ): Promise<void> => {
   const colonyClient = await getCachedColonyClient(colonyAddress);
 
-  const parsedAction = parseAction(action, [colonyClient, null, null, null]) as
+  const parsedAction = parseAction(action, { colonyClient }) as
     | TransactionDescription
     | undefined;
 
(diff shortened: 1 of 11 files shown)


jakubcolony (approved):
LGTM 👍

src/handlers/motions/motionFinalized/helpers.ts
```diff
@@ -375,3 +390,111 @@ export const updateColonyUnclaimedStakes = async (
...
   }
 };
+
+export const claimExpenditurePayouts = async (
+  action: string,
+  colonyAddress: string,
+): Promise<void> => {
+  const colonyClient = await getCachedColonyClient(colonyAddress);
+
+  const parsedAction = parseAction(action, [colonyClient, null, null, null]) as
```
bassgeta, line 400: do we need the 3 nulls here? since the function will break if client doesn't exist
image

bassgeta (approved):
works, looks good, just dropped a comment which may or may not be relevant

src/handlers/motions/motionFinalized/helpers.ts
iamsamgibbs: I’ve changed the type to this:

```
interface MotionActionClients {
  colonyClient?: AnyColonyClient | null;
  oneTxPaymentClient?: AnyOneTxPaymentClient | null;
  stakedExpenditureClient?: AnyStakedExpenditureClient | null;
  stagedExpenditureClient?: AnyStagedExpenditureClient | null;
}
```

Which I prefer as when calling the function it is more explicit which clients are being passed and can choose not to pass particular clients.

I just wanted to confirm that the order in which we check the clients doesn’t matter? The previous type enforced the order so that ColonyClient is checked first before the extensions. I don't think there is, but I’m not 100% sure if there could be a situation where the matching function fragment could exist on multiple clients and whether we could want to prioritise the colony client.

bassgeta: i like this new type, so all good from my side :+1:

bassgeta (approved):

src/handlers/motions/motionFinalized/helpers.ts
jakubcolony: @bassgeta what do you mean by "break"? 

@iamsamgibbs that looks like a good idea, it's now down to the developer to make sure they're passing colony client as the first property of that clients object.

bassgeta: well if client was null it just didn't go through, I didn't double check the type and only later found that we need to always pass an array of length 4 in. But this new type seems clearer

PR #185, opened 2024-03-05
feat: add finalizedAt timestamp if motion finalized via permissions

bassgeta:
Basically what the title says :D


jakubcolony (approved):
Does what it says, nice spot ;)

PR #186, opened 2024-03-05
Fix: Value too large for BigNumber conversion and comparison

iamsamgibbs:


rdig (approved):
Changes seem sensible, thanks for fixing it!

src/handlers/motions/motionFinalized/helpers.ts
```diff
@@ -440,7 +439,7 @@ export const claimExpenditurePayouts = async (
   // - The value is not 0
   // - The storage slot is not 26
   // Then we can assume this state change ain't a stage release
-  if (convertedStorageSlot !== 26 || convertedValue !== 0) {
+  if (convertedStorageSlot !== 26 || value !== BigNumber.from(0)) {
```
jakubcolony, line 442: I don't think this is gonna work as you're not comparing primitives anymore but objects.
I tested it with something like this:
```ts
BigNumber.from(0) === BigNumber.from(0)
```
and got `false`, because they're 2 different objects.

How about:
```ts
BigNumber.from(0).eq(value)
```

jakubcolony (changes requested):

src/handlers/motions/motionFinalized/helpers.ts
iamsamgibbs: Ah yep, good catch, fixed this.

src/handlers/motions/motionFinalized/helpers.ts
```diff
@@ -440,7 +439,7 @@ export const claimExpenditurePayouts = async (
   // - The value is not 0
   // - The storage slot is not 26
   // Then we can assume this state change ain't a stage release
-  if (convertedStorageSlot !== 26 || convertedValue !== 0) {
+  if (convertedStorageSlot !== 26 || BigNumber.from(0).eq(value)) {
```
jakubcolony, line 442: Just a small fix to check the value is _not_ 0:
```suggestion
  if (convertedStorageSlot !== 26 || !BigNumber.from(0).eq(value)) {
```

jakubcolony (changes requested):

src/handlers/motions/motionFinalized/helpers.ts
iamsamgibbs: 🤦‍♂️ Fix is up now!

jakubcolony (approved):
LGTM 👍

PR #189, opened 2024-03-08
Feat: Implement edit locked expenditure motion

iamsamgibbs:
[Related CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/2021)

This PR implements the `editLockedExpenditureMotion` and refactors the motion created multilcall handlers. Previously the handler split up all the multicall actions, checked the fragment name and ran the relevant handler function - this essentially meant all the actions were handled independently.

This created an issue for the `editLockedExpenditureMotion` as all the proposed edits need to be handled together and stored in the motion. This refactor adds a 'validator' function for each multicall handler to determine which motion called the multicall handler and pass it to the appropriate 'handler' function which handles the entire multicall rather than the individual actions as before - enabling actions to be handled in a batch where necessary.


src/constants.ts
```diff
@@ -12,3 +13,15 @@ export const SUPPORTED_EXTENSION_IDS = [
 ];
 
 export const SIMPLE_DECISIONS_ACTION_CODE = '0x12345678';
+
+const toB32 = (input: BigNumberish): string =>
```
rdig, line 17: I remember we had a util for this somewhere, but for the life of me can't find it right now

src/handlers/expenditures/helpers/decodeSetExpenditureState.ts
```diff
@@ -50,7 +46,7 @@ export const decodeUpdatedSlots = (
         recipientAddress,
       });
     } else if (keys[1] === EXPENDITURESLOT_CLAIMDELAY) {
-      const claimDelay = toNumber(value);
+      const claimDelay = toNumber(value).toString();
```
rdig, line 49: I assume we convert these to strings so that we can save them via mutations, right ?

src/handlers/motions/motionCreated/handlers/multicall/fragments.ts
```diff
@@ -1,29 +1,46 @@
...
+    [isReleaseExpenditureStageMotion, releaseExpenditureStageMotionHandler],
+    [isEditLockedExpenditureMotion, editLockedExpenditureMotionHandler],
+  ];
+
+// List all supported multicall fragments
+export const supportedMulticallFragments: string[] = [
+  'moveFundsBetweenPots(uint256,uint256,uint256,uint256,uint256,uint256,address)',
+  'moveFundsBetweenPots(uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,address)',
+  'setExpenditureState',
+  'setExpenditurePayout(uint256,uint256,uint256,uint256,address,uint256)',
```
rdig, line 45: Don't we have enums for these method signatures ?

_(if not, we should)_

src/handlers/motions/motionCreated/handlers/multicall/multicallHandlers/editLockedExpenditureMotion.ts
```diff
@@ -0,0 +1,130 @@
...
+import { MulticallHandler, MulticallValidator } from '../fragments';
+
+export const isEditLockedExpenditureMotion: MulticallValidator = ({
+  decodedFunctions,
+  expenditureStatus,
+}) => {
+  const fragmentsToMatch = [
+    'setExpenditurePayout(uint256,uint256,uint256,uint256,address,uint256)',
+    'setExpenditureState',
+  ];
```
rdig, line 26: Same here about signatures

src/handlers/motions/motionCreated/handlers/multicall/multicallHandlers/editLockedExpenditureMotion.ts
```diff
@@ -0,0 +1,130 @@
...
+  let updatedSlots: ExpenditureSlot[] = [];
+
+  if (!expenditure) {
+    return;
+  }
+
+  for (const decodedFunction of decodedFunctions) {
+    if (
+      decodedFunction.fragment ===
+      'setExpenditurePayout(uint256,uint256,uint256,uint256,address,uint256)'
```
rdig, line 50: signature enum

src/handlers/motions/motionCreated/handlers/multicall/multicallHandlers/editLockedExpenditureMotion.ts
```diff
@@ -0,0 +1,130 @@
...
+
+      updatedSlots = getUpdatedExpenditureSlots(
+        expenditure,
+        convertedSlot,
+        {
+          payouts: updatedPayouts,
+        },
+        updatedSlots,
+      );
+    } else if (decodedFunction.fragment === 'setExpenditureState') {
```
rdig, line 88: enum

src/handlers/motions/motionCreated/handlers/multicall/multicallHandlers/fundExpenditureMotion.ts
```diff
@@ -2,15 +2,47 @@ import { Result } from 'ethers/lib/utils';
...
+import { MulticallHandler, MulticallValidator } from '../fragments';
 
-export const moveFundsBetweenPotsMulti = async ({
+export const isFundExpenditureMotion: MulticallValidator = ({
+  decodedFunctions,
+  expenditureStatus,
+}) => {
+  const fragmentsToMatch = [
+    'moveFundsBetweenPots(uint256,uint256,uint256,uint256,uint256,uint256,address)',
+    'moveFundsBetweenPots(uint256,uint256,uint256,uint256,uint256,uint256,uint256,uint256,address)',
```
rdig, line 21: signature enums

src/handlers/motions/motionCreated/handlers/multicall/multicallHandlers/releaseExpenditureStageMotion.ts
```diff
@@ -0,0 +1,46 @@
...
+import { toNumber } from 'lodash';
+import { getDomainDatabaseId, getExpenditureDatabaseId } from '~utils';
+import { createMotionInDB } from '~handlers/motions/motionCreated/helpers';
+import { MulticallHandler, MulticallValidator } from '../fragments';
+
+export const isReleaseExpenditureStageMotion: MulticallValidator = ({
+  decodedFunctions,
+  expenditureStatus,
+}) => {
+  const fragmentsToMatch = ['setExpenditureState'];
```
rdig, line 11: enum

rdig (changes requested):
All good, I just rejected this so you can convert the method signatures to enums.

Nice work overall!

src/handlers/motions/motionCreated/handlers/multicall/multicallHandlers/editLockedExpenditureMotion.ts
```diff
@@ -0,0 +1,130 @@
...
+        [];
+
+      const networkInverseFee = (await getNetworkInverseFee()) ?? '0';
+      const amountLessFee = getAmountLessFee(
+        amountWithFee,
+        networkInverseFee,
+      ).toString();
+      const feeAmount = BigNumber.from(amountWithFee)
+        .sub(amountLessFee)
+        .toString();
```
jakubcolony, line 66: Would be great to extract this to avoid repetition between handlers

jakubcolony (changes requested):
As mentioned in the other PR, great work overall and a very neat solution for the multicall handlers! 

I'm only requesting changes due to a few suggestions I had, functionally everything works as expected

src/handlers/expenditures/helpers/getUpdatedSlots.ts
```diff
@@ -8,17 +8,37 @@ export const getUpdatedExpenditureSlots = (
   expenditure: ExpenditureFragment,
   slotId: number,
   fieldsToUpdate: Partial<Omit<ExpenditureSlot, 'id' | '__typename'>>,
+  previouslyUpdatedSlots?: ExpenditureFragment['slots'],
```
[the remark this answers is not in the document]
iamsamgibbs: Both are needed when checking for a matching slot in previouslyUpdatedSlots - it is possible that there are previouslyUpdatedSlots, but not with a matching slotId, in which case we grab the existingSlot from the expenditure. So in this instance both the previouslyUpdatedSlots and the expenditure.slots are needed.

I have still refactored so that the first argument only accepts the slots rather than the entire expenditure for the sake of consistency between the two arguments.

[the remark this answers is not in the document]
jakubcolony: Oh sorry, I did not realise that could be the case. In this scenario, is it worth making a separate function for your purposes, since it's specific to updating slots via multicall?

jakubcolony (approved):
Thanks for making the changes @iamsamgibbs, looks good! 👍

jakubcolony (approved):
Top notch! 👌

src/constants.ts
```diff
@@ -12,3 +14,14 @@ export const SUPPORTED_EXTENSION_IDS = [
...
+// These are indexes of data slots in contracts
+// For more information: https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/colony/ColonyStorage.sol
+export const EXPENDITURES_SLOT = BigNumber.from(25);
+export const EXPENDITURESLOTS_SLOT = BigNumber.from(26);
+
+export const EXPENDITURE_OWNER_AND_STATUS = toB32(BigNumber.from(0));
+
+export const EXPENDITURESLOT_RECIPIENT = toB32(BigNumber.from(0));
+export const EXPENDITURESLOT_CLAIMDELAY = toB32(BigNumber.from(1));
+export const EXPENDITURESLOT_PAYOUTMODIFIER = toB32(BigNumber.from(2));
```
rdig, line 27: This is very confusing to read, as we're mixing "magic" numbers (slot indexes)  with actual numbers 

I'd change this to something like:
```js
export const EXPENDITURES_SLOT = BigNumber.from(SlotIndex.Expenditures);
export const EXPENDITURESLOTS_SLOT = BigNumber.from(SlotIndex.ExpenditureSlots);
```

_(Naming sucks because of they network named these)_

rdig (approved):
Left a comment to maybe improve the re ability a bit, but overall I'm good with this.

Very nice work Sam! :100:

PR #190, opened 2024-03-11
Move `rootHash` to action model

ArmandoGraterol:
CDapp PR: https://github.com/JoinColony/colonyCDapp/pull/2035

This PR updates the parts of the block ingestor that were using and inserting the root hash on the motion model, now it should be on the action model so that it's available on motions and actions at the same time.


jakubcolony (approved):
Changes look good, nice work 👍

PR #193, opened 2024-03-27
Finalizing expenditure as non-owner (using motions)

jakubcolony:
https://github.com/JoinColony/colonyCDapp/pull/2097


ArmandoGraterol (approved):
Tested this code over in the CDapp and it's working! Also looks fine to me.

rdig (approved):
All good here as well. Codewise, at least, since the functional test was done via JoinColony/colonyCDapp#2097 's review

iamsamgibbs (approved):
Tested with it's CDapp counterpart - do we need to check colony version here?

PR #195, opened 2024-04-01
Minimise number of RPC calls made while syncing

area:
## Explanation

While exploring the possibility of deploying on Arbitrum, a make-or-break issue was found with the block ingestor:

screenshot-2024-04-01_16-13-08

This is primarily because blocks on Arbitrum are 25x more frequent than they are on Gnosis (five-per-second and one-per-five-seconds, respectively). The block ingestor was able to process 100 blocks per 10-12 seconds; this is enough to keep pace and eventually sync, but if a resync was ever required (which it will be) it would have been essentially impossible.

From initial testing, this PR appears to increase the syncing performance of the block ingestor by a factor of 40 or 50, at least in the regime where most blocks do not contain transactions of interest and we have few if any colonies tracked. This makes syncing much more comfortable. There are some optimisations that will help once those assumptions are not true somewhere on Raul's machine - we could add them in this PR, or later (though today I have learned that Arbitrum blocks [will never be empty](https://www.degencode.com/p/decoding-the-arbitrum-sequencer-feed#%C2%A7l-and-l-messages), so that would only be a benefit for chains where that could be true).

## Approaches taken
The main reason for syncing being as slow as it was is due to the number of RPC requests made, and so I have done my best to remove as many as I can:

### `blockListener.js`
When started, the block ingestor would start `trackMissedBlocks`. This made a `getBlock` call for every block that was missing, until the `blocksMap` was filled with all the blocks that we were going to need. However, at this early point in the ingestion, we don't actually need that information, and we might not need it at all. So instead, we just track the most recent block we've seen (i.e. the block with the biggest number). I realise that this backfilling was non-blocking, but if there's no need to make the requests, we may as well not.

### `blockProcessor`
In `processNextBlock`, we call `getLogs` for the block. But instead, we can call `getLogs` for (as implemented here) up to 1000 blocks, and store them until we need them.

I also have resolved a TODO around `blocksMap`, and delete a block we've finished processing from it.

### `events.ts`
This is the only place we actually need the block hash and timestamp - if we are processing a block that has an event we're interested in. Now we can reasonably call `getBlock` if we need to, and store the result in case we need it later.

## Other notes

In summary, when we are syncing blocks that contain no colony events, this PR moves us from making 2 RPC requests per block to 1/1000 RPC requests per block, which results in a significant speedup. When we are synced, this seems to move to about 1/80 RPC requests per block, which is a function of the polling interval in ethers.

There's a lot of logging around the sync status that I introduced. I'm happy to remove or alter it, but it was extremely useful!

I'm not sure where to make this PR to - either `master-arbitrum` or `master`. I assume we want these improvements on `master` eventually, but for right now I don't know.

I am still a little apprehensive about the improvement seen here, and am worried I've optimised something critical out, but I've tested it locally and it seems to work, and also picked up a colony being created when we ran it through some very limited testing on the Arbitrum Sepolia test deployment.

flex

### DIFF
--- a/src/blockListener.ts
+++ b/src/blockListener.ts
@@ -6,15 +6,19 @@ import { processNextBlock } from '~blockProcessor';
 /**
  * Map storing blocks that have been either picked up by the block listener
  * or missed blocks tracking
- * @TODO: Explore the possiblity of removing blocks once they've been processed
+ * Blocks are removed once processed by a call to .delete in the blockProcessor
  */
 export const blocksMap = new Map<number, Block>();
+let latestSeenBlockNumber = 0;
+
+export const getLatestSeenBlockNumber = (): number => latestSeenBlockNumber;
+// export const blocksMap = new Map<number, boolean|Block>();
 
 export const startBlockListener = (): void => {
   provider.on(EthersObserverEvents.Block, async (blockNumber: number) => {
     try {
-      const block = await provider.getBlock(blockNumber);
-      blocksMap.set(block.number, block);
+      // For now, we just track that this block exists.
+      latestSeenBlockNumber = Math.max(latestSeenBlockNumber, blockNumber);
 
       output(`Block ${blockNumber} added to the queue`);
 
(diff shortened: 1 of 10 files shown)


rdig (approved):
Nicely done. Merging

PR #196, opened 2024-04-04
Advanced Payments API: Return info about Cancel action

jakubcolony:
https://github.com/JoinColony/colonyCDapp/pull/2132


iamsamgibbs (approved):
All works as expected, just need to get rid of those pesky console logs!

PR #197, opened 2024-04-05
Fix: incorrect latest colony version in the DB

jakubcolony:
https://github.com/JoinColony/colonyCDapp/pull/2162

PR #202, opened 2024-04-11
Fix: Incorrect colony version (Arbitrum deployment)

jakubcolony:
This PR applies fix from #197 to `master-arbitrum` branch.


rdig (approved):
So this inadvertently solves a issue we actually with new deployments, that the colony version doesn't get picked up from the contracts (extensions do however), and has to be seeded manually

Nicely done

PR #203, opened 2024-04-11
Refactor: Event listeners

jakubcolony:
This PR refactors the way events are handled in block-ingestor. Previously, after matching an event with an active listener, the event was passed to `eventProcessor` which was effectively a massive `switch` statement over event signature. The downside of this solution was that it was quite difficult to add more than one listener per event signature, even if the filters such as contract address were completely different. This became particularly apparent while discussing the multisig feature, as both extensions share event signatures.

From now, handlers are "attached" to listeners which allows for more granular control over what gets called when. Additionally, the listener gets passed to the handler which made the concept of "additional properties" obsolete.

It's a mid-size refactor, as usual I tried not to make any functional changes outside of the scope. That being said, a lot of handlers were modified in some way.

## Testing

Run create data script. Test the main flows such as colony creation, most popular actions, motions, balances, members etc.


iamsamgibbs (approved):
Seems good to me! All the code changes look sensible and make sense. Tested a bunch of actions in the CDapp, installed some extensions, made a new colony, all seems to be working fine. Good job! 🎉

src/handlers/expenditures/expenditureMadeStaged.ts
```diff
@@ -1,22 +1,21 @@
...
   UpdateExpenditureMutationVariables,
 } from '~graphql';
-import { ContractEvent } from '~types';
-import { getExpenditureDatabaseId, output, toNumber, verbose } from '~utils';
+import { EventHandler } from '~types';
+import { getExpenditureDatabaseId, toNumber, verbose } from '~utils';
 
-export default async (event: ContractEvent): Promise<void> => {
-  const { colonyAddress } = event;
+export const handleExpenditureMadeStaged: EventHandler = async (
```
bassgeta, line 12: not sure how much overhead this would add, but should we type this specific function differently, so it receives an `ExtensionEventListener` in the parameter type definition?

bassgeta: Maybe we could have different event handlers typed and reuse those,
so we have a `ExtensionEventHandler` which gets a listener that's of the type `ExtensionEventListener`

Then we can have a general `EventHandler` type which would just be a union of all different types?

bassgeta (approved):
Looks slick, worked for me when I tried creating some actions and create-data ran over without any issues :tada: 
Just left a comment if we feel like doing it, it would get rid of the `listener as ExtensionEventListener` statements we have in some parts

src/handlers/expenditures/expenditureMadeStaged.ts
jakubcolony: I've tried that initially and battled it tonight but generics in union types defeated me again. 

The problem here is that on one hand, we want to make listeners discriminated by handler type, but each handler can be used by multiple listener types - e.g. `ExpenditureCancelled` can be emitted by either colony or extension.

The most correct approach would be to narrow down the listener type but it feels like a bit of an overkill (you'd need a function, if check, etc.)

PR #204, opened 2024-04-16
[Multi-sig] feat: implement installation and global configuration of the multi-sig extension

bassgeta:
[The CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/2229)

### DIFF
--- a/src/eventProcessor.ts
+++ b/src/eventProcessor.ts
@@ -54,6 +54,7 @@ import {
   handleColonyMetadataDelta,
   handleMotionEventSet,
 } from './handlers';
+import { handleMultiSigGlobalThresholdUpdated } from '~handlers/multiSig';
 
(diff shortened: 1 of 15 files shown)


src/eventProcessor.ts
```diff
@@ -75,6 +77,7 @@ export default async (event: ContractEvent): Promise<void> => {
     );
   }
 
+  output('dobijo ja event brate', event.signature);
```
iamsamgibbs, line 80: Output to remove

iamsamgibbs (approved):
Reviewed at the same time as CDapp#2229 - code looks good here too! Will need to be refactored to match #203 when that is possible.

ArmandoGraterol (approved):
If this isn't good, I don't know what is!

PR #208, opened 2024-04-22
Update default root domain colour

davecreaser:
## Changes

- Update of the generated code following this PR: https://github.com/JoinColony/colonyCDapp/pull/2269
- Change the default colour of the root domain on colony creation

## Testing

- First, ensure that the CDapp PR is merged, or that you're running it all locally with both branches.
- Create a new colony
- Check that the colour of the General domain is blue-400

### DIFF
--- a/src/graphql/generated.ts
+++ b/src/graphql/generated.ts
@@ -1490,6 +1490,20 @@ export type CreateTransactionInput = {
   titleValues?: InputMaybe<Scalars['String']>;
 };
 
+/** Input data for creating a unique Colony within the Colony Network. Use this instead of the automatically generated `CreateColonyInput` input type */
+export type CreateUniqueColonyInput = {
+  /** Unique identifier for the Colony. This is the Colony's contract address */
+  colonyAddress: Scalars['ID'];
+  /** User id of creator to associate with further invite codes */
+  initiatorAddress: Scalars['ID'];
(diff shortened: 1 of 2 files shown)


src/graphql/generated.ts
```diff
@@ -1490,6 +1490,20 @@ export type CreateTransactionInput = {
...
+  /** User id of creator to associate with further invite codes */
+  initiatorAddress: Scalars['ID'];
+  /** Unique identifier for the Colony's native token (this is its address) */
+  tokenAddress: Scalars['ID'];
+  /** The transaction hash of colony creation transaction */
+  transactionHash: Scalars['String'];
+  /** Type of the Colony (regular or MetaColony) */
+  type?: InputMaybe<ColonyType>;
+};
+
```
jakubcolony, line 1506: Similarly to CDapp codegen will need to be run again

jakubcolony (changes requested):

iamsamgibbs (changes requested):
All looks good! Just a codegen re-run as Jakub already mentioned.

iamsamgibbs (approved):
Looks good to me! Those extra deletions look like they're leftover from a previous change and are no longer needed so can be removed here.

jakubcolony (approved):

PR #213, opened 2024-05-01
Optimize Block Processing

rdig:
> ## Disclamer
>
>
>

## Current PR Details

Most, well... rather all work in this PR has been actually done by @area so all kudos go to him.

This change comes as a response to our Arbitrum deployment and it's way faster block speeds _(`250ms` / block)_, which was way more faster than we were equipped to handle _(gnosis has a `5sec` / block interval)_.

To improve the block ingestor to handle these new speeds, Alex made block processing in batches _(configurable via a ENV variable)_, and just handling the relevant logs from that batch. This comes a improvement over old version of processing each block individually, as well as fetching logs for each block, even if they were not relevant to the ingestor.

## Testing

Testing this is kinda tricky since you won't be able to actually get close to those kind of speeds locally, but you can check the output of the ingestor and see the batching in action. You should see a lot of blocks being added to the queue and one batch to process them. Every `BLOCK_PAGING_SIZE` block you should also see a message telling you how long it took to process that batch of blocks

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -11,3 +11,7 @@ AWS_APPSYNC_KEY=da2-fakeApiId123456
 NODE_ENV=development
 
 BLOCK_TIME=5 # in seconds
+
+# Number of blocks to process in a single batch
+# Helps greatly with fast networks (ie: Arbitrum)
+BLOCK_PAGING_SIZE=1000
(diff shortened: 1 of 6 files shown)


iamsamgibbs (approved):
Looks good to me!

Screenshot 2024-05-03 at 22 25 15

PR #214, opened 2024-05-01
Fix Colony Creation (on networks with faster block time)

rdig:
> ## Disclamer
>
>
>

## Current PR Details

This is actually just a fix of the `tryFetchGraphqlQuery` util _(work done by @jakubcolony)_ which is used to "wait" until a colony's metadata exist, then proceed with the creation. This piece of logic was used to separate chain created colonies from UI created colonies.

However, this util was actually broken, but due to the slower speeds of Gnosis, it was never revealed. Only on Arbitrum, with it's faster block times, did this problem actually surface.

## Testing

Testing this is kinda tricky since you won't be able to actually get close to those kind of speeds locally, but just creating a colony should do the trick. If your colony gets created, then you're all good.


iamsamgibbs (approved):
Seems good! Tested and was able to create a new colony.

PR #215, opened 2024-05-01
Fix OneTxPayment Expenditure Balance Not Updating

rdig:
> ## Disclamer
>
>
>

## Current PR Details

All of the work here was done by @jakubcolony so all kudos go to him.

This fixes a issue where an amount being payed out of the colony would decrease the colony's balance for twice that amount.

This was due to a issue where the simple payment's expenditure _(one tx payment's expenditure)_ was not being updated properly, and showing a negative value. 

Combine that with the balance being fetched by lambdas with the payment value already subtracted, and the expenditure balance "added" to the (already subtracted) balance, and presto change-o you have double the amount removed

## Testing

Create a fresh colony [via the UI](https://github.com/JoinColony/colonyCDapp/pull/2308), mint a memorable amount _(`100` or `1000`)_. Create a payment for yet another memorable amount _(`10` or something)_.

Look up the expenditure in the database via the graphql explorer, it should have the balance `0`. Any other value in there means this is still broken. _(for one OneTX Payments, other expenditure types, can have non-zero values)_

Screenshot from 2024-05-01 13-32-14

Screenshot from 2024-05-01 13-33-58

Look up, via the UI](https://github.com/JoinColony/colonyCDapp/pull/2308), the colony's balance table, which should should your initial minted amount, minus the payment amount _(if it shows double the payment amount, then this is still broken)_

Screenshot from 2024-05-01 13-35-23

Contributes to JoinColony/colonyCDapp#2298


iamsamgibbs (approved):
Tested with CDapp#2308

PR #217, opened 2024-05-03
Feat: Cancel streams using permissions

iamsamgibbs:
[See CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/2321)


jakubcolony (changes requested):
Looking great, just a small comment around fetching from the chain after which it should be good to go!

jakubcolony (approved):
LGTM

rdig (approved):
All good

PR #218, opened 2024-05-07
Fix: Fund expenditure motion does not get created in the DB

jakubcolony:
https://github.com/JoinColony/colonyCDapp/pull/2330


rdig (approved):
Tested via JoinColony/colonyCDapp#2330

PR #221, opened 2024-05-10
Fix: Use StaticJsonRpcProvider where appropriate

iamsamgibbs:
[CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/2356)

### DIFF
--- a/src/provider.ts
+++ b/src/provider.ts
@@ -12,7 +12,9 @@ export const setChainId = (newChainId: ChainID): void => {
 };
 export const getChainId = (): ChainID => chainId;
 
-const provider = new providers.JsonRpcProvider(process.env.CHAIN_RPC_ENDPOINT);
+const provider = new providers.StaticJsonRpcProvider(
+  process.env.CHAIN_RPC_ENDPOINT,
+);
 
 export const initialiseProvider = async (): Promise<void> => {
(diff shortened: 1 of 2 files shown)


src/provider.ts
```diff
@@ -12,7 +12,9 @@ export const setChainId = (newChainId: ChainID): void => {
 };
 export const getChainId = (): ChainID => chainId;
 
-const provider = new providers.JsonRpcProvider(process.env.CHAIN_RPC_ENDPOINT);
+const provider = new providers.StaticJsonRpcProvider(
+  process.env.CHAIN_RPC_ENDPOINT,
+);
```
rdig, line 17: I'd want @area 's input here, because the way I think about this, we still need the "normal", "noisy" `JsonRpcProvider` in the block ingestor as opposed to the `StaticJsonRpcProvider` one

rdig (changes requested):
Good with this, but want Alex's input before we merge it

src/provider.ts
area: Chatted briefly with Raul on a call, his worry was around listeners firing, but they should still fire with the StaticJsonRpcProvider.

rdig (approved):
All good now!

jakubcolony (approved):

PR #230, opened 2024-05-30
fix: Allow any wallet address to be a verified member

rumzledz:
### Description

This PR will now allow any wallet addresses to be a Colony contributor which should satisfy this AC: "You should be able to verify addresses, even if they are not members"

It also removes all references to a Colony whitelist and is done as part of [colonyCDapp#2450](https://github.com/JoinColony/colonyCDapp/pull/2450) to address issue [colonyCDapp#2314](https://github.com/JoinColony/colonyCDapp/issues/2314).

### Behaviour

New Contributors added via the Manage Verified Members flow should not be able to view the Colony right away. They should still have to go through the Colony invite flow.

### Merge Considerations

Because this is a dependency for [colonyCDapp#2450](https://github.com/JoinColony/colonyCDapp/pull/2450), I will need to liaise with a fellow dev (most likely @jakubcolony) to make sure that both PRs are merged at pretty much the same time.

### DIFF
--- a/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
+++ b/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
@@ -52,7 +52,6 @@ export const handleAddVerifiedMembers = async (
         await createColonyContributor({
           colonyAddress,
           contributorAddress: userAddress,
-          isWatching: true,
         });
 
         return;
--- a/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
+++ b/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
@@ -52,6 +52,9 @@ export const handleAddVerifiedMembers = async (
         await createColonyContributor({
           colonyAddress,
           contributorAddress: userAddress,
+          hasPermissions: false,
+          isVerified: true,
+          colonyReputationPercentage: 0,
         });
 
         return;
--- a/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
+++ b/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
@@ -52,10 +52,17 @@ export const handleAddVerifiedMembers = async (
         await createColonyContributor({
           colonyAddress,
           contributorAddress: userAddress,
-          isWatching: true,
         });
 
-        return;
+        await mutate<
+          UpdateColonyContributorMutation,
+          UpdateColonyContributorMutationVariables
+        >(UpdateColonyContributorDocument, {
+          input: {
+            id: getColonyContributorId(colonyAddress, userAddress),
+            isWatching: true,
+          },
+        });
       }
 
       // If user is already verified, don't verify them again
--- a/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
+++ b/src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
@@ -52,6 +52,7 @@ export const handleAddVerifiedMembers = async (
         await createColonyContributor({
           colonyAddress,
           contributorAddress: userAddress,
+          isWatching: true,
         });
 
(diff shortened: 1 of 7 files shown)


src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
```diff
@@ -43,6 +46,25 @@ export const handleAddVerifiedMembers = async (
         id: getColonyContributorId(colonyAddress, userAddress),
       });
 
+      // If a wallet address is not currently a member of a Colony,
+      // add the entity as a new Colony contributor
+      if (item?.data?.getColonyContributor === null) {
```
bassgeta, line 51: I think we could do this with one less GQL call, by checking if it's null, then creating a contributor with `isWatching` set to `true` otherwise do the usual flow of updating the contributor document

bassgeta (changes requested):
Loving the negative line count :sunglasses: 
Just dropped a comment regarding creation of the contributor, otherwise stellar job!

src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
rumzledz: > I think we could do this with one less GQL call, by checking if it's null, then creating a contributor with isWatching set to true 

I tried to do that yesterday but the block-ingestor was throwing all sorts of errors I'm not exactly sure why 😕 I deleted the containers, ran my dev and even restarted my laptop! But the same issue kept coming up. Then I did this instead and it was working so I stuck with it.

rumzledz: See this commit @bassgeta https://github.com/JoinColony/block-ingestor/pull/230/commits/7a056883458a945d9245fbd285ce8206f627020a

rumzledz: @bassgeta I'll make sure to run the block-ingestor right after amplify is finished! I've reverted back to the changes I made to set `isWatching` to `true` on Contributor creation!

rumzledz: @bassgeta I have also updated the behaviour for new Contributors who were added via the Manage Verified Members flow. I added a screenshot of the conversation I had with Arren about this. Apologies, can you review again? Thank you! 🙏

src/handlers/metadataDelta/handlers/addVerifiedMembers.ts
```diff
@@ -43,6 +46,17 @@ export const handleAddVerifiedMembers = async (
...
+        await createColonyContributor({
```
jakubcolony, line 52: I don't think you can simply reuse this function.

It creates a contributor, but it sets `isVerified` to false and, potentially worse, `hasPermissions` to true.

jakubcolony (changes requested):

jakubcolony (approved):
LGTM 🚀

rumzledz:
Hey @jakubcolony @bassgeta sorry I had to rebase. Can I please get you to review again? Thank you!

rdig:
@rumzledz a couple of things:
- please don't post internal Discord screenshots on public Github issues
- As for the title, "Non-Colony" is not the correct naming convention, as by definition, all addresses on the block chain are non-colony, including the contracts we deploy

rdig (approved):
All good on the ingestor side as well

bassgeta (approved):
LGTM :sunglasses:

rumzledz:
> @rumzledz a couple of things:
> 
> * please don't post internal Discord screenshots on public Github issues
> * As for the title, "Non-Colony" is not the correct naming convention, as by definition, all addresses on the block chain are non-colony, including the contracts we deploy

All right noted!

PR #231, opened 2024-05-31
Feat: Manage Tokens Refactor

iamsamgibbs:
This is the feature branch for the Manage Tokens Refactor


arrenv (approved):
@iamsamgibbs Tested fully, all good, I noticed some unexpected motion behavior, which I have raised here - https://github.com/JoinColony/colonyCDapp/issues/2713

But, it is non-breaking and I feel we can treat it separately and merge without it.

PR #233, opened 2024-06-04
[Multi-sig] feat: finalization of a multi sig motion

bassgeta:
[CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/2465)
Changes:
* handler for multi sig motion executed
* moved some code around so I could unify action handling from both multi sig and voting reputation
* moved `linkpendingMetadata` helpers to the utils folder, so they can be reused by multi sig and voting reputation (also changed it a bit and made it work with multi sig)
* moved parsing of action to common helper
* moved `createColonyAction` to common helper
* moved `updateDecisionInDB` to a common helper (not used yet)

There's still an issue with calling `linkPendingMetadata` by the motion executed handler, since it should parse an action string, which we don't receive after motion execution. 
We can tackle this while implementing the add domain motion or something :)

Do note I've added an `isSuccess` field to the multi sig entity, that's because of the [contract itself ](https://github.com/JoinColony/colonyNetwork/blob/20dd443bc7e70e00eadc5541400baacb11d1f58f/contracts/extensions/MultisigPermissions.sol#L349)


src/handlers/multiSig/helpers.ts
```diff
@@ -42,3 +53,68 @@ export const getUserMultiSigSignature = async ({
...
+      'Could not find the multiSig in the db. This is a bug and needs investigating.',
+    );
+  }
+
+  return multiSig;
+};
+
+export const updateMultiSigInDB = async (
+  multiSigData: Partial<ColonyMultiSig> & { id: string },
+  showInActionsList?: boolean,
```
jakubcolony, line 81: I dislike that field for its lack of granularity, at the moment actions can be hidden for different reasons:
- motions below staking threshold (hardcoded in block-ingestor)
- if initiator/recipient is an extension
- if it's move funds to expenditure

But it's very hard to tell exactly why. It would be great for MultiSig not to use it.

If there's a use case to hide MultiSigs in UI, it would be preferred to store the data and let the client decide on filtering.

jakubcolony (changes requested):

src/handlers/multiSig/helpers.ts
bassgeta: fair point, I was replicating the way motions work, but honestly Multi Sig can just always show up. Will double check with product, but I think we can just automatically set this to true when it's created.

src/handlers/multiSig/multiSigExecuted/multiSigExecuted.ts
```diff
@@ -0,0 +1,58 @@
...
+  if (finalizedMultiSig) {
+    /* @TODO fix this up when we start porting over domain motions
+     * the action string is not an event argument like it's for motions, so we may need to tweak linkPendingMetadata a bit
+    await linkPendingMetadata(
+      action,
+      colonyAddress,
+      finalizedMultiSig.id,
+      true,
+    );
+    */
```
jakubcolony, line 47: Forgot to comment on this one yesterday, you can potentially get it from contracts or even better, store it in the DB when multisig gets created

bassgeta: thanks for this, I'll copy this comment to an issue which will tackle this. Sadly I currently do not have any way of testing if it will work because those motions do not have multi sig support yet

iamsamgibbs (approved):
Looks good to me!

Tested alongside CDapp#2465

jakubcolony (approved):
LGTM

PR #238, opened 2024-06-10
Fix: ingestor possibly skipping some events

jakubcolony:
It looks like by mistake, we never awaited promises inside `setupListenersForExtensions` which meant the event processing would potentially start before the listeners were in place.

## Steps to reproduce the original issue
I noticed this when testing https://github.com/JoinColony/colonyCDapp/pull/2490.

- Create a motion
- Disable block-ingestor (it will need to be run outside Docker at that point)
- Stake/oppose the motion
- Enable block-ingestor

In `master`, the stake event would not get picked up by ingestor and get lost without updating the motion. This fix should rectify that issue.

### DIFF
--- a/src/eventListeners/extension/index.ts
+++ b/src/eventListeners/extension/index.ts
@@ -86,11 +86,11 @@ export const setupListenersForExtensions = async (): Promise<void> => {
     ),
   );
 
-  setupListenerForOneTxPaymentExtensions();
-  setupListenersForVotingReputationExtensions();
-  setupListenersForStakedExpenditureExtensions();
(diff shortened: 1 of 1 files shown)


iamsamgibbs (approved):
Good catch and a nice simple fix.

Tested and the stake support was picked up upon restarting the block-ingestor:

Screenshot 2024-06-11 at 10 45 26

Screenshot 2024-06-11 at 10 48 25

rumzledz (approved):
LGTM ✅

rdig (approved):
Nice catch

davecreaser (approved):

PR #246, opened 2024-06-21
Feat: Handle both addDomain and editDomain actions created via a Multi-Sig decision method

rumzledz:
This is needed to test [colonyCDapp#2588: Feat: enable addDomain multi-sig flow](https://github.com/JoinColony/colonyCDapp/pull/2588)

### DIFF
--- a/src/handlers/multiSig/multiSigCreated/handlers/addOrEditDomain.ts
+++ b/src/handlers/multiSig/multiSigCreated/handlers/addOrEditDomain.ts
@@ -14,7 +14,7 @@ export const handleAddOrEditDomainMultiSig = async (
 
   const { name, args: actionArgs } = parsedAction;
 
-  const domainId = actionArgs[5];
+  const domainId = actionArgs[2];
 
   const fromDomainId = getDomainDatabaseId(colonyAddress, domainId);
 
--- a/src/handlers/multiSig/multiSigCreated/handlers/addOrEditDomain.ts
+++ b/src/handlers/multiSig/multiSigCreated/handlers/addOrEditDomain.ts
@@ -1,8 +1,9 @@
 import { TransactionDescription } from 'ethers/lib/utils';
 import { ContractEvent, multiSigNameMapping } from '~types';
 import { createMultiSigInDB } from '../helpers';
+import { getPendingMetadataDatabaseId } from '~utils';
 
-export const handleAddDomainMultiSig = async (
+export const handleAddOrEditDomainMultiSig = async (
   colonyAddress: string,
   event: ContractEvent,
   parsedAction: TransactionDescription,
@@ -13,7 +14,13 @@ export const handleAddDomainMultiSig = async (
 
   const { name } = parsedAction;
 
+  const pendingDomainMetadataId = getPendingMetadataDatabaseId(
+    colonyAddress,
+    event.transactionHash,
+  );
+
   await createMultiSigInDB(colonyAddress, event, {
     type: multiSigNameMapping[name],
+    pendingDomainMetadataId,
   });
 };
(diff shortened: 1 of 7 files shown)


bassgeta (changes requested):
Codewise it all looks good, we can get rid of that TODO comment too.
Good job, I've left some comments on the CDapp PR which should be resolved before we get this one in.

src/handlers/multiSig/multiSigExecuted/multiSigExecuted.ts
```diff
@@ -35,6 +37,29 @@ export const handleMultiSigMotionExecuted: EventHandler = async (
...
+    }
+
+    await linkPendingMetadata(
+      actionData,
+      colonyAddress,
+      finalizedMultiSig.id,
+      true,
+    );
+
     /* @TODO fix this up when we start porting over domain motions
```
iamsamgibbs, line 63: This comment can go now that you've implemented it

src/handlers/multiSig/multiSigCreated/handlers/addOrEditDomain.ts
```diff
@@ -0,0 +1,26 @@
...
+  const { name } = parsedAction;
+
+  const pendingDomainMetadataId = getPendingMetadataDatabaseId(
+    colonyAddress,
+    event.transactionHash,
+  );
+
+  await createMultiSigInDB(colonyAddress, event, {
+    type: multiSigNameMapping[name],
+    pendingDomainMetadataId,
```
iamsamgibbs, line 24: We might need to add this to show the `CreatedIn` field on the completed action screen:

```ts
const { name, args } = parsedAction;
const [, , nativeDomainId] = args;

...

fromDomainId: getDomainDatabaseId(colonyAddress, nativeDomainId),
```

(I could be wrong, but it is there for the `editDomain` reputation motion)

iamsamgibbs (changes requested):
Looks good to me and is working as intended. Just one comment to remove, and one little bit to investigate whilst you fix up the CDapp stuff!

src/handlers/multiSig/multiSigCreated/handlers/addOrEditDomain.ts
rumzledz: Hey @iamsamgibbs ! At first I tried to do that before I realised that Kerry has already "pre-added" this inside the `multiSigCreated/helpers.ts` file

Screenshot 2024-07-03 at 14 37 44

src/handlers/multiSig/multiSigExecuted/multiSigExecuted.ts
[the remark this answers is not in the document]
rumzledz: Aye!

rumzledz: Aye!

bassgeta (approved):
Tested the CDapp and everything works, so let's get this merged!

mmioana (approved):
Awesome @rumzledz 👏 

Even if the block ingestor doesn't treat createDomain and editDomain differently, would it make sense to have two separate handlers?

rumzledz:
> Awesome @rumzledz 👏
> 
> Even if the block ingestor doesn't treat createDomain and editDomain differently, would it make sense to have two separate handlers?

Hey @mmioana it does at first glance but after a dozen conversations about this whole code split thing, it's not worth it. The cdapp case is different and so I did the code split there 👌

iamsamgibbs (approved):

bassgeta (approved):
Reapproving since reviewed as part of its CDapp counterpart :+1:

rdig (approved):
Tested as part of JoinColony/colonyCDapp#2588

Nicely done once again

PR #249, opened 2024-07-04
Fix: Manage verified users for multi-sig

mmioana:
## Description

Create and finalize multi-sig add and remove verified members motions.

- [X] The add verified members multi-sig motion can be created
- [X] The remove verified members multi-sig motion can be created
- [X] The add verified members motion can be finalized
- [X] The remove verified members motion can be finalized
- [X] Only users with [adequate permissions](https://www.notion.so/colony/Multi-sig-Permissions-6aaf0b95309f4754a243d54e8d9c4189?pvs=4#fec3699a5fd241a2925bd2809494b157) can create them

[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/2650)

### DIFF
--- a/src/handlers/multiSig/multiSigCreated/multiSigCreated.ts
+++ b/src/handlers/multiSig/multiSigCreated/multiSigCreated.ts
@@ -10,11 +10,11 @@ import {
   verbose,
 } from '~utils';
 import {
+  handleMetadataDeltaMultiSig,
   handleMintTokensMultiSig,
   handleUnlockTokenMultiSig,
   handleAddOrEditDomainMultiSig,
 } from './handlers';
-import { handleMetadataDeltaMultiSig } from './handlers/metadataDelta';
 
 export const handleMultiSigMotionCreated: EventHandler = async (
   event,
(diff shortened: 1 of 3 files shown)


src/handlers/multiSig/multiSigCreated/multiSigCreated.ts
```diff
@@ -13,6 +13,7 @@ import {
   handleMintTokensMultiSig,
   handleUnlockTokenMultiSig,
 } from './handlers';
+import { handleMetadataDeltaMultiSig } from './handlers/metadataDelta';
```
bassgeta, line 16: maybe we should import it from the index file?

bassgeta (changes requested):
Tested as part of [CDapp](https://github.com/JoinColony/colonyCDapp/pull/2650), works fine, dropped a few comments there and a minor import related one here.
Nice and straightforward changes!

iamsamgibbs (approved):
Looks great to me!

bassgeta (approved):
Tested as part of [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/2650/files) and it works as expected, nice one!

PR #251, opened 2024-07-04
Add move funds multisig handler

davecreaser:
Part of: https://github.com/JoinColony/colonyCDapp/pull/2657


iamsamgibbs (changes requested):
Change look good here, although it needs a rebase so is hard to tell exactly which files have actually changed. Requesting changes until it has been rebased and until the CDapp counterpart is fixed up. :)

bassgeta (approved):
Tested as part of its [CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/2657), works like a charm, let's gooooo!

rdig (approved):
Tested as part of the review of JoinColony/colonyCDapp#2657

PR #256, opened 2024-07-17
Feat: Edit streams using motions 

iamsamgibbs:
[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/2721)

### DIFF
--- a/src/graphql/fragments/expenditures.graphql
+++ b/src/graphql/fragments/expenditures.graphql
@@ -40,3 +40,45 @@ fragment ExpenditureSlot on ExpenditureSlot {
     networkFee
   }
 }
+
+fragment StreamingPayment on StreamingPayment {
+  id
+  startTime
+  endTime
+  tokenAddress
+  amount
+  interval
+  metadata {
+    ...StreamingPaymentMetadata
+  }
+  claims {
+    amount
+    timestamp
+  }
+  changelog {
+    transactionHash
+    oldValues {
+      ...StreamingPaymentChanges
+    }
+    newValues {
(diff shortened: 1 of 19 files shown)


jakubcolony (changes requested):
Pretty much flawless, top notch work @iamsamgibbs.

I saw a new action type for Edit Streaming Payment added in the schema but can't see it being created anywhere. This would come handy for storing the changes on the action field as explained in the CDapp comment. I think there might be some friction when it comes to cancelling (as we'd need to be careful to only create a cancel action), but feel free to ping me if anything comes up

iamsamgibbs:
> Pretty much flawless, top notch work @iamsamgibbs.
> 
> I saw a new action type for Edit Streaming Payment added in the schema but can't see it being created anywhere. This would come handy for storing the changes on the action field as explained in the CDapp comment. I think there might be some friction when it comes to cancelling (as we'd need to be careful to only create a cancel action), but feel free to ping me if anything comes up

Thanks Jakub! There is already a check in `endTimeSet` to check if the new endTime is equal to the current blocktime and if so assume that this is a cancel action, so hopefully this is already accounted for, but I'll give it a check!

rumzledz (approved):
Nothing stands out to me and everything is working as expected in the cdapp PR. Awesome stuff @iamsamgibbs ✨

jakubcolony (approved):
LGTM 👍

jakubcolony (approved):

mmioana (approved):
Such a pleasure to look over the code @iamsamgibbs 🤩

PR #259, opened 2024-07-30
Fix: conditionally add multisig chunk to historic role id

rumzledz:
This PR fixes an issue with how we build a historic role's ID.

Historic roles IDs will now conditionally have the "_multisig" chunk when a multi-sig role is being processed.

Please test this as part of: [colonyCDapp#2773: Auto-populate the Role and Permissions fields in the Manage Permissions form ](https://github.com/JoinColony/colonyCDapp/pull/2773)


bassgeta (approved):
Nice and simple :ok_hand: Tested the CDapp counterpart, let's merge this badboi!

iamsamgibbs (approved):
Thanks for adding this! Tested with the CDapp counterpart.

PR #261, opened 2024-08-05
feat: remove gas estimation for motions

bassgeta:
This PR removes all gas estimation from the block ingestor in favor of calling different contracts as defined in its [CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/2817)


rdig (approved):
So much cleaner now that `gasEstimate` doesn't need to be passed around everywhere. 

I've also tested functionally as part of the review of JoinColony/colonyCDapp#2817

Nicely done again Kerry

iamsamgibbs (approved):
👋 Good riddance gasEstimate!

mmioana (approved):
Awesome @bassgeta for removing the `gasEstimate`
Tested as part of the [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/2817)

PR #266, opened 2024-08-12
fix: create historic roles for motions too, fix check if roles are multi-sig condition when creating a multi-sig motion

bassgeta:
[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/2924)

### DIFF
--- a/src/handlers/motions/motionCreated/handlers/setUserRoles.ts
+++ b/src/handlers/motions/motionCreated/handlers/setUserRoles.ts
@@ -3,12 +3,19 @@ import { BigNumber, constants } from 'ethers';
 
 import { ContractEvent, motionNameMapping } from '~types';
 import {
+  createColonyHistoricRoleDatabaseEntry,
   getColonyRolesDatabaseId,
   getDomainDatabaseId,
   getRolesMapFromHexString,
 } from '~utils';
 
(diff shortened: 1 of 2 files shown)


src/handlers/motions/motionCreated/handlers/setUserRoles.ts
```diff
@@ -32,6 +41,38 @@ export const handleSetUserRolesMotion = async (
...
+      domainId.toNumber(),
+      userAddress,
+      blockNumber,
+      {
+        ...existingRoles,
+        ...roles,
+      },
+      isMultiSig,
+    );
+  }
```
mmioana, line 74: Given we use the same piece of code in both `src/handlers/motions/motionCreated/handlers/setUserRoles.ts` and `src/handlers/multiSig/multiSigCreated/handlers/setUserRoles.ts`, could we extract it in a helper function?

mmioana (approved):
Tested as part of the [colonyCDapp PR](https://github.com/JoinColony/colonyCDapp/pull/2924)
I left a small comment, if you feel life refactoring a tiny bit 👀

rdig (approved):
All good. Tested this functionally while doing the review of JoinColony/colonyCDapp#2924

PR #268, opened 2024-09-03
Send notifications on simple payment action

davecreaser:
Part of CDapp PR https://github.com/JoinColony/colonyCDapp/pull/3055

### DIFF
--- a/src/utils/notifications.ts
+++ b/src/utils/notifications.ts
@@ -1,0 +1,144 @@
+import MagicBellClient, { Notification } from '@magicbell/core';
+import { query } from '~amplifyClient';
+import {
+  GetColonyContributorsNotificationDataDocument,
+  GetColonyContributorsNotificationDataQuery,
+  GetColonyContributorsNotificationDataQueryVariables,
+  NotificationsDataFragment,
+} from '~graphql';
+
+export enum NotificationType {
+  SimplePayment = 'SimplePayment',
+  Mention = 'Mention',
+}
+
+interface NotificationVariables {
+  title: string;
+  creator: string;
+  colonyAddress: string;
+  transactionHash: string;
+  notificationType: NotificationType;
+  amount?: string;
+  tokenAddress?: string;
+  mentions?: string[];
+}
+
+// Set up the notification client
+export const setupNotificationsClient = async (): Promise<void> => {
+  await MagicBellClient.createInstance({
+    apiKey: process.env.MAGICBELL_API_KEY as string,
+    apiSecret: process.env.MAGICBELL_API_SECRET as string,
+  });
+};
+
+// Check if notifications should be sent to a recipient from a colony.
+const shouldSendNotificationToRecipient = (
+  notificationsData: NotificationsDataFragment,
+  colonyAddress: string,
+): boolean => {
+  return (
+    !!notificationsData.magicbellUserId && // User has a magicbell user account made
+    !notificationsData.notificationsDisabled && // User has not disabled notifications app-wide
+    !notificationsData.mutedColonyAddresses.includes(colonyAddress) // User has not muted the colony that the notification is in
+  );
+};
+
+// Get all the recipients of a colony wide notification in the format Magicbell expects.
+const getRecipientsOfColonyWideNotification = async (
+  colonyAddress: string,
+): Promise<Array<{ external_id: string }>> => {
+  const recipients: Array<{ external_id: string }> = [];
+  let nextToken: string | null | undefined;
+
+  do {
+    const members = await query<
+      GetColonyContributorsNotificationDataQuery,
+      GetColonyContributorsNotificationDataQueryVariables
+    >(GetColonyContributorsNotificationDataDocument, {
+      colonyAddress,
+      limit: 100,
+      ...(nextToken ? { nextToken } : {}),
+    });
+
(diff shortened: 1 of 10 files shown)


iamsamgibbs (approved):
Tested with colonyCDapp#3055

iamsamgibbs:
Oh! Forgot to mention, can you add the new .env variables to example.env ?

src/utils/notifications.ts
```diff
@@ -0,0 +1,144 @@
...
+    } else {
+      nextToken = null;
+    }
+  } while (nextToken);
+
+  return recipients;
+};
+
+// Send a notification when an action is made.
+export const sendActionNotification = async ({
```
bassgeta, line 90: should we maybe split this into a 
`sendColonyWideNotification` which will send it to all recipients (we can filter out the mentions array if that type of notifications exists)
`sendActionNotification` which will send notifications directly for the impacted user addresses? Then we could also query notification data just for those users instead of always getting notification data for all colony contributors?

bassgeta (changes requested):
Good work, sensible code changes and looks pretty easy to build upon this.
I am wondering if my comments are applicable, maybe I'm missing some context here, but I feel like we will mostly be sending out notifications colony wide or for a few members, so it could be beneficial to split those up so we don't always fetch the entire contributors list.
It would also be nice if we would have a muted flag stored on the colony contributor, so we could immediately filter over it, but I do understand it may be a bit problematic with the notifications ids and stuff stored on the user object

src/utils/notifications.ts
```diff
@@ -0,0 +1,144 @@
...
+): boolean => {
+  return (
+    !!notificationsData.magicbellUserId && // User has a magicbell user account made
+    !notificationsData.notificationsDisabled && // User has not disabled notifications app-wide
+    !notificationsData.mutedColonyAddresses.includes(colonyAddress) // User has not muted the colony that the notification is in
+  );
+};
+
+// Get all the recipients of a colony wide notification in the format Magicbell expects.
+const getRecipientsOfColonyWideNotification = async (
```
mmioana, line 47: Maybe it would make sense to group a bit the functionality here
So my proposal would be to first get all paginated members data
```
interface GetDataArgs extends Record<string, unknown> {
  nextToken?: string | null;
  colonyAddress: string;
}

interface GetDataResponse<TItem> {
  items: (TItem | null)[] | undefined,
  nextToken?: string | null;
}

type GetDataFn<T> = (args: GetDataArgs) => Promise<GetDataResponse<T> | null | undefined>

const getMembersData = async ({ colonyAddress, nextToken }: GetDataArgs): Promise<GetColonyContributorsNotificationDataQuery['getContributorsByColony']> => {
  const response = await query<
      GetColonyContributorsNotificationDataQuery,
      GetColonyContributorsNotificationDataQueryVariables
    >(GetColonyContributorsNotificationDataDocument, {
      colonyAddress,
      limit: 100,
      ...(nextToken ? { nextToken } : {}),
    })

    return response?.data?.getContributorsByColony
}

const getAllPages = async <T>(getData: GetDataFn<T>, params: GetDataArgs) => {
  const items = [];
  let nextToken = null;

  do {
    const data = await getData({ ...params, nextToken });
    nextToken = data?.nextToken;
    if (data?.items) {
      items.push(...data.items);
    }
  } while (nextToken);

  return items;
};

```

and then use it in this function as

```
const members = await getAllPages(getMembersData, { colonyAddress })

members.forEach(...)
```

src/utils/notifications.ts
```diff
@@ -0,0 +1,144 @@
...
+    recipients,
+    custom_attributes: {
+      notificationType,
+      creator,
+      colonyAddress,
+      transactionHash,
+      ...(amount ? { amount } : {}),
+      ...(tokenAddress ? { tokenAddress } : {}),
+    },
+  });
```
mmioana, line 119: Maybe we should add a `try ... catch` block here

mmioana (changes requested):
Awesome work @davecreaser and very clean and straight-forward code 💯 

I left some suggestions, if you would mind considering them

Tested as part of the [CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/3055)

src/utils/notifications.ts
davecreaser: So actually on every single action we will be sending a colony wide notification, but the mention will be optional, which is why I structured it this way. I have created a new `createNotification` function which can now be used to just send a straight up notification if we need to do that at all though, skipping all the contributor fetching steps.

davecreaser:
@bassgeta and @mmioana - changes have been made and it's ready to review again :) I've also made changes on the CDapp counterpart, and it should now work to run the block ingestor in docker, you'll just need to ensure that your CDapp .env file includes the magicbell api key and secret!

bassgeta (approved):
Nice changes, love the new util we can use for just sending flat out `n` notifications.
Looks legit, works legit :ship: ! 
Tested as part of [its CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/3055)

mmioana (approved):
Thank you for implementing the changes @davecreaser 💯 
Everything looks nice and tidy 👏 
Tested as part of the [CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/3055#pullrequestreview-2317563450)

PR #269, opened 2024-09-03
Prevent adding Incoming Funds Claims from Non Supported Tokens

rdig:
Prior to this fix, anything "acting" like a "valid" ERC-20 tokens could send any colony some amount of tokens.

Granted, the UI would have been filtering out the incoming transfer, if said token was not on the "approved" tokens list _(although you could still see it manipulating the filters)_

Now this is where the problem gets complicated. Let's say I create a contract that capable of transferring token amounts to other addresses _(a Colony in our example)_, but one which doesn't meet the [ERC-20 standard specification](https://ethereum.org/en/developers/docs/standards/tokens/erc-20/) _(as low of a bar as that is)_. If such a token would make a transfer to an existing Colony, the transfer would be recorded, and the UI will fetch it as "valid" incoming funds, while simultaneous failing to fetch the token's data since the Lambda logic would prevent it. 

This would end up leaving a colony unable to load, due to the token data not being found

Screenshot from 2024-09-03 17-08-58

This now creates an attack vector. Knowing this I can "fairly" easily craft a new contract that simulates a token, and emits `Transfer` events, and just send some tokens to one of the known colonies, instantly causing service disruption.

You can even take this one step further. Deploy multiple such contracts, and send all chain colonies _(this data is public)_ some tokens, each colony being "funded" from a different contract. That way it would make it even more difficult to detect.

The fix however is fairly strait forward, as we just need to ensure any tokens that we process a incoming funds claim from, needs to pass our token validation logic. If it doesn't then there's no reason to add the claim to the colony's bookkeeping.

A word of **warning** however, that even with this fix, a crafty developer can write a contract to correctly "pass" all our checks _(they're public after all)_, and then have some other contract call broken or doing nefarious things. This will not cause an exact service disruption, as the colony will still load, but will block other parts of the app, especially the ones that deal with interacting with a token.

### Testing

Testing this is rather problematic as it's very hard to simulate in a local environment. I will be describing the steps that are required to achieve this, but I cannot provide a concrete contract for your to actually test.

So, to test this, one would have to create a new token contract, but one that doesn't have one of the methods we check for _(in this case `allowance` would be a good choice to remove)_. Write that contract, compile it and deploy it to your local dev environment via `hardhat`. 

Once you've done that, instantiate it, and using the newly created Token instance, mint some tokens, unlock them _(if the contract has such functionality)_ and send them to a colony.

Since the token is not "valid" from a CDapp perspective, your incoming funds transfer will not be picked up by the ingestor.

### DIFF
--- a/src/handlers/colonies/transfer.ts
+++ b/src/handlers/colonies/transfer.ts
@@ -82,12 +82,14 @@ export default async (event: ContractEvent): Promise<void> => {
         : '',
     );
 
+    let tokenAddressFromDb: string | undefined;
+
     /**
      * Call the GetTokenFromEverywhere query to ensure the token
      * gets added to the DB if it doesn't already exist
      */
     try {
-      await query<
+      const dbTokenQuery = await query<
         GetTokenFromEverywhereQuery,
         GetTokenFromEverywhereQueryVariables
       >(GetTokenFromEverywhereDocument, {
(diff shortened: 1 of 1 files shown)


iamsamgibbs (approved):
Nice work figuring this issue out. I've not attempted to test this, but the code changes to confirm the token passes the GetTokenFromEverywhere "tests" before creating the claim makes sense as an attempt to reduce this vulnerability.

Happy to test this if anyone provides a concrete step by step guide, but equally happy to just approve this without testing based on the code changes.

rdig:
So this came about since we just had a case in prod, and this fix is intended to prevent further such incidents.

I will however endeavor to create that "broken token" contract so that we can test locally. 

I might be fun after all :)

rdig:
@iamsamgibbs and whoever else would like to test this fix, you can find testing instructions here JoinColony/colonyCDapp#3077

Note that they need to be run on the CDapp side rather then here

iamsamgibbs (approved):
Tested as part of colonyCDapp#3077

mmioana (approved):
Cool and sensitive change @rdig 👍 

Tested as part of [#3077](https://github.com/JoinColony/colonyCDapp/pull/3077)

PR #272, opened 2024-09-30
Send notifications on permissions actions

davecreaser:
See description and testing steps in CDapp PR: https://github.com/JoinColony/colonyCDapp/pull/3205

### DIFF
--- a/src/handlers/actions/colonyUpgrade.ts
+++ b/src/handlers/actions/colonyUpgrade.ts
@@ -7,9 +7,10 @@ import {
 } from '~graphql';
 import { ContractEvent } from '~types';
 import { toNumber, verbose, writeActionFromEvent } from '~utils';
+import { sendActionNotifications } from '~utils/notifications';
 
 export default async (event: ContractEvent): Promise<void> => {
-  const { contractAddress: colonyAddress } = event;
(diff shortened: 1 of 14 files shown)


bassgeta (approved):
Approving since I've tested its CDapp counterpart, looking good and simple!

mmioana (approved):
Really nice and simple solution @davecreaser 🌟 
Tested as part of the [colonyCDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3205#pullrequestreview-2347464579)

@davecreaser I have just a curiosity, is there a reason for which we don't `await` the `sendActionNotifications` calls?

iamsamgibbs (approved):
Tested with the CDapp counterpart, these all look like sensible changes. Nice job!

davecreaser:
> Really nice and simple solution @davecreaser 🌟 Tested as part of the [colonyCDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3205#pullrequestreview-2347464579)
> 
> @davecreaser I have just a curiosity, is there a reason for which we don't `await` the `sendActionNotifications` calls?

Yeah, It's basically just so they aren't blocking anything else. It's not crucial that they are sent immediately, so they shouldn't block other, more important things and instead should just be completed whenever the resources are available :)

PR #276, opened 2024-10-08
[Notifications] feat: send notifications when new colony version added to network

bassgeta:


src/handlers/colonies/colonyVersionAdded.ts
```diff
@@ -2,12 +2,21 @@ import { COLONY_CURRENT_VERSION_KEY } from '~constants';
...
-  await updateCurrentVersion(COLONY_CURRENT_VERSION_KEY, convertedVersion);
+  const handleVersionUpdated = async (): Promise<void> => {
+    await sendColonyVersionAddedNotifications(convertedVersion);
+  };
+
+  await updateCurrentVersion(
+    COLONY_CURRENT_VERSION_KEY,
+    convertedVersion,
+    handleVersionUpdated,
+  );
```
jakubcolony, line 21: is there a benefit to such callback approach over simply calling it after `updateCurrentVersion`?

src/utils/notifications.ts
```diff
@@ -390,6 +394,50 @@ export const sendFundsClaimedNotifications = async ({
...
+
+      if (recipients.length > 0) {
+        await sendNotification(
+          `New colony version: ${newVersion}`,
+          recipients,
+          {
+            notificationType: NotificationType.NewColonyVersion,
+            notificationCategory: NotificationCategory.Mention,
+            colonyAddress: colony.id,
+            creator: colony.id, // who's the creator of this??
```
jakubcolony, line 432: good question 😄

jakubcolony (approved):
LGTM, as tested with https://github.com/JoinColony/colonyCDapp/pull/3283

src/handlers/colonies/colonyVersionAdded.ts
bassgeta: well come to think of it, mostly local dev stuff, `updateCurrentVersion` updates the version only if the "new" version is higher than the current one so I've opted for some old school callbacks :D 
Lemme know if this won't ever trigger, I can just remove it in that case and immediately invoke the function instead of passing it over
image

src/utils/notifications.ts
davecreaser: 🤯  is the creator field optional? Maybe it's about time that it is!

davecreaser: Otherwise yeah just the colony id is fine 😆

src/handlers/colonies/colonyVersionAdded.ts
davecreaser: How about just calling `sendColonyVersionAddedNotifications` directly inside the `updateCurrentVersion` util?

davecreaser (approved):
Nice work @bassgeta, all looking good. Tested with the CDapp part of this PR and all is working.

rdig (changes requested):
Reject just because I can't test JoinColony/colonyCDapp#3283 properly

iamsamgibbs (approved):
Reviewed alongside the CDapp counterpart, nice job with the new fragment

src/handlers/colonies/colonyVersionAdded.ts
jakubcolony: It won't ever happen in prod or QA

> How about just calling sendColonyVersionAddedNotifications directly inside the updateCurrentVersion util?

Sounds a bit like SRP violation. The util is also used in multiple places, for different kinds of version updates

bassgeta: will it happen with extensions though? because we use the same helper for updating the extension versions.
It is kinda an SRP violation, but at the same time `currentVersion` is a single entity in the database :shrug: 
Lemme know if we want to refactor this, I can do it on the extensions PR so we wrap this one up and rebase the feature branch

rdig (approved):
Tested as part of the review of JoinColony/colonyCDapp#3283

Everything is working as expected. Nicely done!

PR #278, opened 2024-10-10
[Notifications] Feat: Extensions notifications

iamsamgibbs:
[See CDapp counterpart](https://github.com/JoinColony/colonyCDapp/pull/3290)

(Note: Also updated package-lock.json as it was breaking CDapp, but should probably be done on the feat/in-app-notifications branch)

### DIFF
--- a/src/utils/notifications.ts
+++ b/src/utils/notifications.ts
@@ -17,6 +17,7 @@ export enum NotificationCategory {
   Mention = 'Mention',
   Payment = 'Payment',
   Admin = 'Admin',
+  Extension = 'Extension',
 }
 
 export enum NotificationType {
@@ -41,6 +42,13 @@ export enum NotificationType {
 
   // Actions made with permissions
   PermissionsAction = 'PermissionsAction',
+
+  // Extensions
+  ExtensionInstalled = 'ExtensionInstalled',
+  ExtensionUpgraded = 'ExtensionUpgraded',
+  ExtensionEnabled = 'ExtensionEnabled',
+  ExtensionDeprecated = 'ExtensionDeprecated',
+  ExtensionUninstalled = 'ExtensionUninstalled',
 }
 
 interface NotificationVariables {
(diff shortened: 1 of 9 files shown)


rdig (approved):
Tested as part of the review of JoinColony/colonyCDapp#3290

src/utils/notifications.ts
```diff
@@ -200,6 +223,32 @@ export const sendExpenditureUpdateNotifications = async ({
...
+
+  if (!recipients.length) {
+    return;
+  }
+
+  // Send the colony wide notifications.
+  await sendNotification(`Extension: ${extensionHash}`, recipients, {
+    notificationType,
+    creator,
+    notificationCategory: NotificationCategory.Extension,
```
bassgeta, line 245: cc @davecreaser should we use extension as a category or are these product wise included in the `Admin` category?

bassgeta (approved):
Love the new `updateExtension` helper, it gets the job done well.
And naturally, the entire solution is also very sensible and nice :sunglasses: 
Tested as part of the CDapp :+1:

src/utils/notifications.ts
```diff
@@ -16,6 +16,7 @@ export enum NotificationCategory {
   Mention = 'Mention',
   Payment = 'Payment',
   Admin = 'Admin',
+  Extension = 'Extension',
```
davecreaser, line 19: The categories should be limited to only Mention, Payment and Admin, so all the extension notifications can be under Admin category. Sorry, I should really have left a comment about this when making this type.

src/utils/notifications.ts
davecreaser: Yeah you're correct, they should all use Admin as the category. My fault really for not making that clear in tickets / with comments in the code.

davecreaser: PS. good catch!

rumzledz (approved):

PR #279, opened 2024-10-10
[Notifications] feat: do not send category notifications if a user has them disabled 

bassgeta:
[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3289)

### DIFF
--- a/src/utils/notifications.ts
+++ b/src/utils/notifications.ts
@@ -68,6 +68,17 @@ interface Recipient {
   external_id: string;
 }
 
+const categoryToApiKey: Record<
+  NotificationCategory,
+  | 'adminNotificationsDisabled'
+  | 'mentionNotificationsDisabled'
+  | 'paymentNotificationsDisabled'
+> = {
+  [NotificationCategory.Admin]: 'adminNotificationsDisabled',
+  [NotificationCategory.Mention]: 'mentionNotificationsDisabled',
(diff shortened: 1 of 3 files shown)


rdig (approved):
All good here as well. Tested as part of the review of JoinColony/colonyCDapp#3289

iamsamgibbs (changes requested):
Left a change request on the CDapp counterpart. On a look here, I don't think the issue is caused here, but will leave it as request changes just in case.

src/utils/notifications.ts
davecreaser, line 1: I think the function for sending a permissions action notification is where the source of the error Sam and I mentioned in the CDapp PR is.

There's a line there that says if there are no recipients then return from the function. This happens BEFORE the call to send mentions. I think in our testing, since Leela was the only recipient and we had turned off the notification types for Payments / Admin, that recipients array was empty so it never got to the code to send a mention. Just a theory but seems likely!

davecreaser (changes requested):
Requesting changes for the same reason as Sam, all else looks good, and works good (tested with the CDapp PR).

Super happy with how all the notifications code is being written so far, so great job there!

Just left one comment about where I think the lack of mentions is coming from.

davecreaser (approved):
Approving now since it's working as expected. Good stuff!

iamsamgibbs (approved):

PR #283, opened 2024-10-18
Fix: Only fire one notification on multisig approval

davecreaser:
When a multisig is fully approved, we should ensure that only one notification is fired about this approval. Previously one notification was being sent for each role the user voted with, which in the case of a simple payment for example is three, so you would end up with three notifications all at the same time saying that the multisig motion was approved.

CDapp PR here https://github.com/JoinColony/colonyCDapp/pull/3396 has the testing steps.

### DIFF
--- a/src/handlers/multiSig/approvalChanged/approvalChanged.ts
+++ b/src/handlers/multiSig/approvalChanged/approvalChanged.ts
@@ -1,3 +1,4 @@
+import { BigNumber } from 'ethers';
 import { ExtensionEventListener } from '~eventListeners';
 import { MultiSigVote, NotificationType } from '~graphql';
 import { getChainId } from '~provider';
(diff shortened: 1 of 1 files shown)


bassgeta (approved):
It's always nice to solve an issue in 41 lines when you initially think it's gonna be more like 400 and a week of tears :D 
Really nice solution, super legit to be able to just nick this from the extension contract :100:

src/handlers/multiSig/approvalChanged/approvalChanged.ts
```diff
@@ -66,7 +67,45 @@ export const handleMultiSigApprovalChanged: EventHandler = async (
...
+      BigNumber.from(motion.requiredPermissions).gte(
+        BigNumber.from(1).shl(roleIndex),
+      )
+    ) {
+      // If the permission is required by this multi sig motion...
+      // (this logic is taken from the network, it's very cool I know)
+      if (
+        BigNumber.from(motion.requiredPermissions)
+          .and(BigNumber.from(1).shl(roleIndex))
+          .gt(BigNumber.from(0))
```
mmioana, line 84: Loving the bitwise operations 🥇

mmioana, line 84: Just wondering if it makes sense to extract `BigNumber.from(motion.requiredPermissions)` into a `const`?

mmioana (approved):
Great work on this one @davecreaser 👏 💯 

Tested as part of the [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3396#pullrequestreview-2381453190)

Nortsova (approved):
Tested it in scope of https://github.com/JoinColony/colonyCDapp/pull/3396 .

Nicely done @davecreaser 🙌

PR #284, opened 2024-10-18
Feat: Add payout claimed notification

davecreaser:
Add a new notification for expenditures when a payout is claimed.

CDapp PR here: https://github.com/JoinColony/colonyCDapp/pull/3400

### DIFF
--- a/src/handlers/expenditures/expenditurePayoutClaimed.ts
+++ b/src/handlers/expenditures/expenditurePayoutClaimed.ts
@@ -9,6 +9,7 @@ import {
 } from '~utils';
 import {
   ExpenditurePayout,
+  NotificationType,
   UpdateExpenditureDocument,
   UpdateExpenditureMutation,
   UpdateExpenditureMutationVariables,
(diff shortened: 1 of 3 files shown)


bassgeta (approved):
Nice and clean addition :fire: 
Tested as part of its CDapp counterpart and can confirm it works!

mmioana (approved):
Nice and straight-forward change 💯 👏 

Tested as part of the [colonyCDapp](https://github.com/JoinColony/colonyCDapp/pull/3400#pullrequestreview-2381377584)

src/handlers/expenditures/expenditurePayoutClaimed.ts
```diff
@@ -25,6 +27,7 @@ export default async (event: ContractEvent): Promise<void> => {
     slot,
     token: tokenAddress,
     tokenPayout: amountWithoutFee,
+    agent: initiatorAddress,
```
jakubcolony, line 30: Just checking that you want this notification to be sent to the user who claimed the payout, and not the recipient/owner of the expenditure?

jakubcolony (approved):
LGTM, tested with colonyCDapp#3400.

src/handlers/expenditures/expenditurePayoutClaimed.ts
davecreaser: It should just be sent colony wide, it's not a specific notification. The initiator address will be used to get information on who triggered the notification (creator), but in reality is isn't really used for anything other than debugging.

PR #289, opened 2024-10-28
fix: always use roles passed to contract for motions

bassgeta:
Instead of getting the "delta" state and applying just the difference in roles in our database, we should just use whatever the FE passed to the contracts and apply it.
[CDApp PR](https://github.com/JoinColony/colonyCDapp/pull/3490)

### DIFF
--- a/src/handlers/multiSig/multiSigCreated/handlers/setUserRoles.ts
+++ b/src/handlers/multiSig/multiSigCreated/handlers/setUserRoles.ts
@@ -4,6 +4,7 @@ import { ContractEvent, multiSigNameMapping } from '~types';
 import { getDomainDatabaseId, getRolesMapFromHexString } from '~utils';
 import { createMultiSigInDB } from '../helpers';
 import { sendMentionNotifications } from '~utils/notifications';
+import { constants } from 'ethers';
 
 export const handleSetUserRolesMultiSig = async (
   colonyAddress: string,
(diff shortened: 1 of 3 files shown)


iamsamgibbs (approved):
Reviewed with the CDapp counterpart. All working as expected and nice simplification of the code here now.

mmioana (approved):
Nice and much easier to follow through with your changes @bassgeta Tested as part of the [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3490#pullrequestreview-2401501784)

rdig (approved):
Tested as part of JoinColony/colonyCDapp#3490 

All working as intended. Nicely fixed :100:

src/handlers/multiSig/multiSigCreated/handlers/setUserRoles.ts
```diff
@@ -20,22 +17,12 @@ export const handleSetUserRolesMultiSig = async (
   }
   // When setting 'Own' authority, the action target will be the colonyAddress
   // When setting 'Multisig' authority, the action target will be the multisig extension address
-  const isMultiSig = actionTarget !== colonyAddress;
+  const isMultiSig = actionTarget !== constants.AddressZero;
```
iamsamgibbs, line 20: Oh, good catch! Can you update the comment here too?

PR #290, opened 2024-10-29
feat: advanced payments funding step multisig

bassgeta:
Sorry about the line count, it's reformatting of `generated.ts` :shrug: 
[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3553)
Changes:
- Split up `multiSigCreated` flow, now we handle "multiple functions" differently like we do with multicall. This is just the base for future advanced payment work
- Added the `fundExpenditure` handler for multi-sig, which works quite similar to how the motion multicall does
- Multisig `moveFunds` handler now assumes that if an expenditure is tied to this action, we should call `fundExpenditure` instead
- `writeAction` now checks if the action was motion finalization for multi-sig too

### DIFF
--- a/src/utils/parseFunction.ts
+++ b/src/utils/parseFunction.ts
@@ -1,5 +1,6 @@
 import { utils } from 'ethers';
 import { TransactionDescription } from 'ethers/lib/utils';
+import { output } from './logger';
 
 /**
  * Helper attempting to decode function data by trying to parse it with different contract ABIs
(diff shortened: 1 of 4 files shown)


src/handlers/multiSig/multiSigCreated/multiSigCreated.ts
```diff
@@ -76,9 +67,29 @@ export const handleMultiSigMotionCreated: EventHandler = async (
...
+
+  const actions = motion.data;
+  const actionTargets = motion.targets;
+
+  if (!actions[0]) {
+    verbose(`No action data in multiSig motion: ${motionId}`);
+    return;
+  }
+
+  console.log('hendlamo single function a');
```
iamsamgibbs, line 82: Leftover console log

iamsamgibbs (changes requested):
Rejecting due to the changes requested on the CDapp, everything (except the console log) looks good here though!

jakubcolony (approved):
Great work @bassgeta, tested with CDapp counterpart and all seems to be working well 👍

src/utils/parseFunction.ts
mmioana, line 1: 🙌

mmioana (approved):
Re-tested as part of the [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3553#pullrequestreview-2462252688)

PR #291, opened 2024-10-31
Fix: Edit expenditure updating payouts when token changes

jakubcolony:
A subtle change without which setting new token for a payout while also editing an existing one did not work.

Consider the following scenario:

The UI wants to change a payout token from Token A to Token B. Since payouts can have multiple tokens, this process requires two steps: setting the new amount for Token B and setting Token A's amount to 0.

block-ingestor picks up the first `ExpenditurePayoutSet` for Token B and adds the updated payout to existing slot. However, when processing the second event, instead of applying the update to the recently modified slot, it incorrectly applies it to the original, unmodified slot. This results in the loss of Token B's payout information.


rdig (approved):
Nice find, and fix! :+1:

iamsamgibbs (approved):
Looks like a sensible change to me!

PR #296, opened 2024-11-22
Fix Reputation Mining Cycle Handler Timestamp

rdig:
This PR fixes the timestamp value for the `reputationMiningCycle` handler, since if the ingestor falls behind the chain, by the time it catches back up to this event it will have used the wrong timestamp for reputation, meaning both updates and UI reporting will be wrong.

This value is mostly used for UI display purposes, however, even so, if it gets out of date, users might get the wrong impression when looking up the reputation cycle's next update in the user hub

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "hasInstallScript": true,
       "license": "GPL-3.0",
       "dependencies": {
-        "@colony/colony-js": "7.2.0",
+        "@colony/colony-js": "^8.0.0-next.0",
         "@colony/events": "3.0.0",
         "@magicbell/core": "^5.0.16",
         "aws-amplify": "^4.3.43",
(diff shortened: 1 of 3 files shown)


iamsamgibbs (approved):
Reviewed alongside the CDapp counterpart.

jakubcolony (approved):
Tested alongside [colonyCDapp#3756](https://github.com/JoinColony/colonyCDapp/pull/3756) 👍

bassgeta (approved):
Damn this was definitely a digging session, props for this solution :muscle: 
Before applying the patch:
image
After:
image

nice one :sunglasses:

rdig:
@bassgeta can I get the "corresponding" review on the CDapp side too please :)

rdig:
This is on hold until the may deployment goes out, will merge after

mmioana (approved):
Reviewed this as part of the [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3756#pullrequestreview-2506172840)

PR #297, opened 2024-11-28
[Proxy colonies] feat: refactor block ingestor into pnpm workspaces 

bassgeta:
This seems like a lot, but let me assure you that most of it is actually just the addition of `pnpm-lock.yaml` :D 

#### New
- we now use `pnpm` instead of `npm` 
- pnpm workspaces :tada: 
- Modified the Dockerfile to build the new app via pnpm workspaces

#### Modified
- the current block ingestor was moved into an app called `main-chain`, the code is mostly intact bar the graphql imports
- the graphql part was decoupled into a shared package (the first one :tada: )
- loading the environment is now done via a single script which we call in the `package.json` file before running the ingestor


apps/main-chain/src/index.ts
```diff
@@ -14,6 +14,10 @@ import { setupNotificationsClient } from '~utils/notifications';
 
 utils.Logger.setLogLevel(utils.Logger.levels.ERROR);
 
+//@TODO remove these after testing
+console.log(`Magicbell dev key var`, process.env.MAGICBELL_DEV_KEY)
+console.log(`RPC endpoint var`, process.env.CHAIN_RPC_ENDPOINT)
```
iamsamgibbs, line 19: 👀

davecreaser (approved):
Really nice stuff guys @bassgeta and @mmioana!

I was initially terrified of the number of changes but you're totally right, it's not as big and scary as it seems 😅 

I've had a look through it all and nothing stands out to me. Really like the packages / apps layout 👍 

I'll give it my seal of approval! 🦭

iamsamgibbs (approved):
Reviewed alongside the CDapp counterpart

apps/main-chain/src/utils/notifications.ts
```diff
@@ -303,7 +303,7 @@ export const sendMentionNotifications = async ({
 
   const validRecipients: Recipient[] = [];
 
-  (response?.data?.listUsers?.items ?? []).forEach((user) => {
+  (response?.data?.listUsers?.items ?? []).forEach((user: any) => {
```
jakubcolony, line 306: Was this a temporary override or is it still needed?

jakubcolony (approved):
Kudos to both of you, this is looking like a beginning of a very nice monorepo setup! 👍 

It works perfectly when running it as part of the dev env, but I had some issues when trying to run ingestor separately. I assume at least some (if not all) can be blamed on me not following some steps, so let me know if that's the case.

I checked out your branch, ran `pnpm install` and then `pnpm run dev:main-chain`. However, I got a bunch of errors that suggest some env values are not read properly (I still had an old `.env` from master at the root directory):

image

----

Some scripts defined in package.json no longer work (I tried both npm run and pnpm run):

`npm run dev`: 

image

`npm run lint`:

image

----

My eslint extension was complaining at the start of every `.ts` file. I had to update `.eslintrc.json`'s `parserOptions` field to point at the new tsconfig.base but I'm not sure if it's the right solution or whether there should be multiple eslint config files per subpackage.
image


----

I'll leave it to your discretion if you want to handle these here or through separate issues and approve this as the dev env works all fine. If you need me to retest any of the stuff above, just give me a shout!

.env.example
```diff
@@ -1,24 +0,0 @@
-STATS_PORT=10001
```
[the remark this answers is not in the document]
bassgeta: yeah the way we made it work (will be more apparent in later PRs) is that we have a common `.env` at the top and then app specific env variables inside each app directory.

bassgeta:
> Kudos to both of you, this is looking like a beginning of a very nice monorepo setup! 👍
> 
> It works perfectly when running it as part of the dev env, but I had some issues when trying to run ingestor separately. I assume at least some (if not all) can be blamed on me not following some steps, so let me know if that's the case.
> 
> I checked out your branch, ran `pnpm install` and then `pnpm run dev:main-chain`. However, I got a bunch of errors that suggest some env values are not read properly (I still had an old `.env` from master at the root directory):
> image
> 
> Some scripts defined in package.json no longer work (I tried both npm run and pnpm run):
> 
> `npm run dev`:
> image
> 
> `npm run lint`:
> image
> 
> My eslint extension was complaining at the start of every `.ts` file. I had to update `.eslintrc.json`'s `parserOptions` field to point at the new tsconfig.base but I'm not sure if it's the right solution or whether there should be multiple eslint config files per subpackage. image
> 
> I'll leave it to your discretion if you want to handle these here or through separate issues and approve this as the dev env works all fine. If you need me to retest any of the stuff above, just give me a shout!

Right since we just wanna merge this asap, I've created a [new issue](https://github.com/JoinColony/colonyCDapp/issues/3924) with minor points which we can tackle whenever and keep these chonkers going into the feature branch.
Thanks a bunch for the review, and do not worry, all of this will be addressed down the line :muscle:

PR #298, opened 2024-12-02
[Proxy colonies] Feat: create proxy chain block ingestor

bassgeta:
This PR adds the proxy block ingestor pnpm app into the workspace that will be responsible for processing the proxy colonies related events.

[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3816)

This is basically reinventing NestJS at this point :sweat_smile: 
The line changes seem absolutely severe, but it's a lot of just moving existing code around, but turning it into classes, and imports, so it's a chonker but not that bad!

We had instances of everything done functional before, which worked fine for a single ingestor, but once we started working on the second one, we quickly realized that we're gonna need dependency injection.
For example, `eventListener` just added handlers to an array, but in order to use it in two apps, we needed to pull it out into a separate package, then instantiate it on each app separately. 
The `blockListener` previously just directly used the running `eventListener` instance. Now the `blockListener` has been turned into a reusable class called `BlockManager` which relies on a renamed `eventListener`: `eventManager`. How to solve all of this? 
Create the `BlockManager` with an instance of `eventManager`, so we can have different instances of `BlockManager` which all just receive `eventManager` instances from each app.
This way, each app just creates its own `eventManager` and then the `BlockManager` with its instance of `eventManager`. Then the `BlockManager` won't care about app proprietary logic, it will just call functions on that instance, getting back data specific to the app it's running in!

### Changes
- Dockerfile stuff to make proxy ingestors buildable
- decoupled amplifyClient, networkClient and rpcProvider into reusable classes we use across both apps
- decoupled the `blockListener` and `blockProcessor` into a new entity called `BlockManager` which does the same things as the two before, additionaly it does event matching (so it's easier to call that)
- decoupled the `eventListener` into a new class, so we can share instances of it across a running chain
- decoupled stats too (there's duplication for the stats server, but that's alright for now because the proxy ingestor requires a really basic setup)
- realigned all imports :melting_face: 
- realigned some codesnips like using `amplifyClient.mutate` instead of just `mutate`


apps/main-chain/src/handlers/colonies/colonyAdded.ts
```diff
@@ -50,8 +50,9 @@ export default async (event: ContractEvent): Promise<void> => {
   /*
    * Add it to the Set
    */
+  // @NOTE seems to not be working, is it borken on master too?
```
iamsamgibbs, line 53: Seemed to work okay for me!

Before creating a new colony (tracked colonies 2)

Screenshot 2024-12-11 at 17 07 46
 
After (tracked colonies 3!)

Screenshot 2024-12-11 at 17 09 33

iamsamgibbs (approved):
Great job on this! Reviewed alongside the CDapp counterpart.

mmioana (approved):
We worked together on it, so can do nothing more but approve ✅

PR #302, opened 2024-12-10
Feat: Add `willPunishExpenditureStaker` field to colony motion

jakubcolony:
Small PR to unblock [colonyCDapp#3322](https://github.com/JoinColony/colonyCDapp/pull/3322) which needs to be able to tell if a motion cancelling staked expenditure will also punish its creator or not.

## Testing 

- Check out `feat/cancel-exp-motion-field` in CDapp and start your dev env as usual.
- Install and enable Voting Reputation
- Create an expenditure and progress it past the Review stage
- From the three dot menu, select the cancel option
- Select Reputation as decision method and try both cases: penalising and not penalising the expenditure creator

Run the following query to check the value of `willPunishExpenditureStaker` field:
```gql

query Motions {
  listColonyActions(filter: {type: {eq: CANCEL_EXPENDITURE_MOTION}}) {
    items {
      id
      motionData {
        willPunishExpenditureStaker
      }
    }
  }
}

```

### DIFF
--- a/src/graphql/generated.ts
+++ b/src/graphql/generated.ts
@@ -1106,6 +1106,11 @@ export type ColonyMotion = {
   /** A list of all of the votes cast within in the motion */
   voterRecord: Array<VoterRecord>;
   voterRewards?: Maybe<ModelVoterRewardsHistoryConnection>;
+  /**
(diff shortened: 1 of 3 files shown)


rdig (approved):
Changes look sensible. Good to go.

PR #304, opened 2024-12-12
feat: handle arbitrary transaction motion

Nortsova:
This PR is cover https://github.com/JoinColony/colonyCDapp/issues/3536  and could be tested in scope of https://github.com/JoinColony/colonyCDapp/pull/3933

### DIFF
--- a/src/handlers/actions/makeArbitraryTransaction.ts
+++ b/src/handlers/actions/makeArbitraryTransaction.ts
@@ -52,7 +52,6 @@ export default async (event: ContractEvent): Promise<void> => {
   await writeActionFromEvent(event, colonyAddress, {
     type: ColonyActionType.MakeArbitraryTransaction,
     initiatorAddress: receipt.from,
-    recipientAddress: receipt.to,
     fromDomainId: getDomainDatabaseId(colonyAddress, Id.RootDomain),
     arbitraryTransactions: [currentArbitraryTransaction],
   });
--- a/src/handlers/motions/motionCreated/handlers/makeArbitraryTransactions.ts
+++ b/src/handlers/motions/motionCreated/handlers/makeArbitraryTransactions.ts
@@ -14,8 +14,7 @@ export const handleMakeArbitraryTransactionsMotion = async (
   event: ContractEvent,
   parsedAction: TransactionDescription,
 ): Promise<void> => {
-  const { name, args: actionArgs } = parsedAction;
-  const [recipients] = actionArgs;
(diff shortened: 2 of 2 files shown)


mmioana (approved):
Nice and simple changes @Nortsova 🙌 
Tested as part of the [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3933#pullrequestreview-2517081371)

bassgeta (approved):
Tested as part of its CDapp counterpart :ok_hand:

src/handlers/actions/makeArbitraryTransaction.ts
```diff
@@ -52,7 +52,6 @@ export default async (event: ContractEvent): Promise<void> => {
   await writeActionFromEvent(event, colonyAddress, {
     type: ColonyActionType.MakeArbitraryTransaction,
     initiatorAddress: receipt.from,
-    recipientAddress: receipt.to,
     fromDomainId: getDomainDatabaseId(colonyAddress, Id.RootDomain),
```
iamsamgibbs, line 55: I'm not sure we actually need this as it is only possible to create arbitrary transactions in the root domain, although looking at some of the other actions we use this `fromDomainId` even if can only be created in the root domain so maybe it is still required.

iamsamgibbs (approved):
Nice job! Reviewed alongside the CDapp counterpart.

src/handlers/actions/makeArbitraryTransaction.ts
jakubcolony: That's a good point, there's no concept of domain when it comes to arbitrary transactions. I would opt for removing this to avoid confusion.

src/handlers/actions/makeArbitraryTransaction.ts
```diff
@@ -52,7 +52,6 @@ export default async (event: ContractEvent): Promise<void> => {
   await writeActionFromEvent(event, colonyAddress, {
     type: ColonyActionType.MakeArbitraryTransaction,
     initiatorAddress: receipt.from,
-    recipientAddress: receipt.to,
```
jakubcolony, line 55: I see what you mean now, it makes sense to remove this as it would always point to the colony rather than the target contract address of the arbitrary transaction.

jakubcolony (approved):
Smashing work @Nortsova! Apart from the Created in field mentioned by Sam, everything looks & works great 💯 

image

image

image

PR #307, opened 2024-12-18
[Arbitrary Txs] Add wagmi, use temporary colony interface 

jakubcolony:
In order to bump colony network version to one that adds an `agent` argument to `ArbitraryTransaction` event, a temporary colony interface is used for parsing logs. This interface is created from wagmi-generated colony ABIs.


jakubcolony:
Closing in favour of #314

PR #312, opened 2025-01-09
Feat: Allow addition of proxy colonies with multisig

davecreaser:


rumzledz (approved):
Tested as part of https://github.com/JoinColony/colonyCDapp/pull/4061. Can confirm that it works as expected! 💯

bassgeta (approved):
Tested the CDapp PR and everything works fine and dandy :sunglasses: 
I wonder who was the buffoon who updated a motion instead of a multisig motion for linking the finalization action :zany_face: mea culpa, thanks for solving it!

PR #314, opened 2025-01-13
[Arbitrary Txs] Get initiator address from event instead of receipt (+ wagmi setup)

jakubcolony:
[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3980)

This PR relies on an updated network version that allows us to grab the arbitrary transaction initiator from the event argument instead of the receipt, which would not work for metatransactions.

To make it work without a new colony-js release, @chmanie kindly helped to set up wagmi CLI to generate ABIs that can be later used to create custom interfaces.


Nortsova (approved):
Great work, thank you 👏 

Tested in scope of https://github.com/JoinColony/colonyCDapp/pull/3980

mmioana (approved):
Really nice work @jakubcolony 🥇 
Thanks for doing the `wagmi` setup as well 🙌 

Everything works as expected and tested it as part of the [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/3980#pullrequestreview-2561809937)

bassgeta (approved):
Tested as part of its CDapp counterpart and can confirm it works :sunglasses: 
Nice one :muscle:

PR #319, opened 2025-01-22
[Proxy colonies M1] Feat: Dynamically add supported proxy chain

mmioana:
Upon startup, add the proxy chain to the DB and mark it as inactive when we shut down the ingestor.
[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/4161)


bassgeta (approved):
Worked on it togetehr and tested as part of its CDapp counterpart :muscle:

iamsamgibbs (approved):
Tested with the CDapp counterpart and all working as expected.

PR #320, opened 2025-01-22
Add rate limited notifications queue

davecreaser:
We recently saw an error where we had hit the API rate limit for Magicbell. This limit is set to 60 requests per minute. There is a decent chance we will hit this limit again - especially in the way that it happened today.

This was when a new version of an extension was released, and we looped over every colony to send a notification to their users who had the permissions to upgrade. We make one API call per colony (luckily we don't need to do it per user since we can pass an array of recipients), and if there are more than 60 colonies as we have on QA, then this will immediately hit our rate limit and the notifications will never be sent.

Even worse than this, it then blocks _any_ notifications from being sent to _anyone_ for potentially the next minute (or until the minute interval has ended) and these notifications are not currently being queued, they just will never fire.

This PR aims to remedy that, by making use of the [p-queue](https://github.com/sindresorhus/p-queue) package. I did some research before picking this package, and find that it is well maintained, well documented, and has a well sized user base. It also fits the needs perfectly for fixing this issue. The caveat is, I had to install it at a slightly earlier-than-latest version, since the latest is ESM only and wasn't compatible with how our block ingestor is set up.

Now, when we want to fire a notification, it enters a queue. The p-queue package handles this queueing and function firing for us, and it will only trigger functions in the queue if we are inside the rate that we have defined (matching Magicbell's 60 per minute).

Now we can feasibly trigger as many notifications at one go as we want, and we won't get rate limited, although it will take some time to fire them all. Better than them all failing after 60 though!

CDapp PR here:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,8 @@
         "express": "^4.18.2",
         "fs-extra": "^10.1.0",
         "graphql": "^16.6.0",
-        "lodash": "^4.17.21"
+        "lodash": "^4.17.21",
+        "p-queue": "^6.6.2"
       },
       "devDependencies": {
         "@colony/wagmi-plugin": "^0.1.1",
@@ -18348,7 +18349,6 @@
     "node_modules/p-finally": {
       "version": "1.0.0",
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=4"
       }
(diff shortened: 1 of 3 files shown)


mmioana (approved):
Really nice and simple solution @davecreaser 👏 

Reviewed as part of the [CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/4162#pullrequestreview-2569191368)

I left there a comment that I still managed to get the error of the `API limit reached` after creating multiple Mint tokens actions.

Also, I am wondering what happens with the queued notifications in case the block-ingestor restarts? 
From what I saw here [p-queue docs](https://www.npmjs.com/package/p-queue)
_For servers, you probably want a Redis-backed [job queue](https://github.com/sindresorhus/awesome-nodejs#job-queues) instead._
So for the future we might want to consider using another queue-ing mechanism if we want to persist the notifications that have not yet been sent 🤔 

Either way I'll approve this PR. Nice work! 💯

davecreaser:
@mmioana - The other rate limits should be fixed now! Left a comment on the CDapp PR, thanks 😄 

And yes, if the block ingestor restarts then the queue will be cleared. I think this is okay for us for now, since notifications are not _crucial_, they're more of a quality of life thing. Also it seems unlikely that we will ever build up much of a queue, apart from the times when a new extension version is released, or a new colony version is released. I did also look into the Redis backed job queue but in my opinion the extra overhead is not worth it, especially at this stage.

bassgeta (approved):
Nice and clean :ok_hand: Tested as part of its CDapp counterpart and can confirm the rate is indeed limited

rdig (approved):
Tested as part of JoinColony/colonyCDapp#4162

Nicely done!

PR #324, opened 2025-01-28
[Proxy Colonies M2] feat: handle when a proxy colony's funds have been claimed

mmioana:
The purpose of this PR is to add a new event listener for the `ProxyColonyFundsClaimed(uint256,address,uint256)` network event, so that once funds on a proxy chain have been claimed, we update them accordingly in the DB.

The `handleProxyColonyFundsClaimed` is pretty similar with the existing `colonyFundsClaimed`, but it might change in the future, depending how we'll want to treat notifications on a proxy chain and if there are any changes needed for the chain native token addresses.

[CDapp PR](https://github.com/JoinColony/colonyCDapp/pull/4200)


apps/main-chain/src/handlers/proxyColonies/proxyColonyFundsClaimed.ts
```diff
@@ -0,0 +1,104 @@
...
+    _amount: amount,
+  } = event.args;
+
+  const chainIdValue = chainId.toNumber();
+
+  // @TODO this needs to be updated as the native chain token claims will have a token address different than constants.AddressZero
+  /*
+   * We're not handling native chain token claims from here, so no point
+   * in running through the whole logic just to end up with the same result
+   */
```
iamsamgibbs, line 32: I _think_ the native chain token will always have address 0x0 so this won't need updating.

My understanding is: It is true that the native chain token on the main chain will be 0x0 and on the other chains it may have a different address, but the native chain token **for each chain** will always be 0x0.

So when it comes to handling these claims here this logic works fine.

Later on, I think we want to handle "linking" tokens in the UI (i.e. CREDS on the main chain is "0x013232..." and on the proxy chain it is "0x01987..." but we want to combine the balances to show it as though it is the same token. The same will be true of ETH on the main chain being "0x0" but on the proxy chain it might be "0x0982...") but I think that will all be handled elsewhere and won't require any changes here.

apps/main-chain/src/handlers/proxyColonies/proxyColonyFundsClaimed.ts
```diff
@@ -0,0 +1,104 @@
...
+      await Promise.all(
+        unclaimedFunds.filter(notNull).map(async ({ id, token, amount }) => {
+          await amplifyClient.mutate<
+            UpdateColonyFundsClaimMutation,
+            UpdateColonyFundsClaimMutationVariables
+          >(UpdateColonyFundsClaimDocument, {
+            input: { id, isClaimed: true },
+          });
+
+          // @TODO this will need to be uncommented and check for the initiator address as we don't get this information from the event
```
iamsamgibbs, line 76: Ah, we will need network to adjust the event to include the initiator address.

iamsamgibbs (approved):
🎉  Reviewed alongside the CDapp counterpart

rdig (approved):
Tested as part of the review of JoinColony/colonyCDapp#4200

Very nice work!

PR #330, opened 2025-02-03
fix: use blockchain time for 'createdAt' field for stakes

Nortsova:
In the UserHub, stakes were created with a real date instead of a blockchain date. So here is the fix for it 🙌


iamsamgibbs (approved):
Reviewed alongside CDapp counterpart.

Makes sense we should be using block time here.

davecreaser (approved):
Cool stuff, reviewed as part of CDapp PR ✅

rdig (approved):
Reviewed as part of JoinColony/colonyCDapp#4238

